Issue #14, opened 2015-08-27
Upgrade to react-bootstrap 0.25


OAGr:
https://www.npmjs.com/package/react-bootstrap

OAGr:
Removed React-Bootstrap, using semantic UI

PR #15, opened 2015-08-27
Upgrad react-bootstrap to 0.25.0

OAGr:
This addresses issue https://github.com/getguesstimate/guesstimate-app/issues/14


OAGr:
This also seems to address the issue that the dropdown wasn't going away after being used :)

PR #35, opened 2015-11-06
Fix typo in README

Jonahss:
^.^


OAGr:
Thanks Jonah!  Though github doesn't seem to have figured out exactly what the difference was.

Issue #39, opened 2015-12-31
uniform distributions have spike at left end-point


masonium:
When I put in a distribution such as [0.05, 0.09] as uniform, There's a large spike in the visualization at the right-most bin.

OAGr:
This is an issue with the histogram library we're using.  If anyone has D3 experience, it would be most appreciated here.

PR #53, opened 2016-01-05
Histogram fix

haryshwaran:
fix for issue #39


OAGr:
Great work @haryshwaran !

On the downside, now this means that the visualization doesn't show the bounds.  I think we should figure out a different way of doing this though.  

image

PR #64, opened 2016-01-08
Fix local installation issues

haryshwaran:
Added dependencies to package.json and cleaned out semantic UI folder.


OAGr:
More description:

Semantic UI has its own build system that comes with it, using a gulp file.  This wasn't fun for use with Guesstimate.  I ended up just using the default Semantic CSS file and then overriding it elsewhere in the codebase.

This was a hurdle for people trying to get Guesstimate set up on their machines.  It didn't seem worth it.  This deletes all of the custom Semantic CSS work, just leaves the regular CSS and JS files.  Right now we don't use the JS files, so we'll probably eventually get rid of them.

Issue #45, opened 2016-01-04
All models sent & in memory


OAGr:
Right now the server returns all models (called spaces in the code) at once.  This is now about 2.5mb.  Super slow and unnecessary. Should be refined.

OAGr:
The great thing about loading in memory is that it makes everything really fast once things are loaded.  The bad thing is that it takes forever to initially load.  Progressive loading solutions can be tricky.  Going to list data that needs to get loaded for different areas:

1) Initial Algolia query should return 30-50 responses, not 15.  Loading more is annoying, there are lots now.

2) No server response should be necessary for page load.  Just algolia.  Right now we use Algolia to find the space IDs, then the in-memory store to show them, which is redundant as Algolia gives us the full entries. 

3) A space should obviously load /space/ID.

We can keep on loading all the users for now, that data is much more minimal.

In the absolute worst case, we could simply just load the space on the space page, and then load the spaces index without the graphs elsewhere.  This is probably a decent first step, should be pretty easy.

OAGr:
This is my #1 priority at this point.  Site is failing for a bunch of people, and may be down soon if we don't act quickly.  Just a heads.

![image](https://cloud.githubusercontent.com/assets/377065/12176473/6772ebcc-b51c-11e5-9fce-ba368ee58609.png)

OAGr:
Thanks to the PR above, this seems fixed. 

Was very interesting (for me).  Basically, it seems like Rails auto cached responses.  As /users and /spaces were called (this loaded everything in them), long-term memory use would rise. The quick rise after the dip at around 10pm was me doing a small bit of load testing on these urls.   I also upgraded the database and reduced the amount it auto saves (in a good way, it used to try after each key stroke in numeric input, making the entire app slower).

Still some more work here, but I think this isn't a bottleneck any more (for now).

![image](https://cloud.githubusercontent.com/assets/377065/12215038/b60143da-b65f-11e5-8741-90edfd0ede66.png)

PR #68, opened 2016-01-09
Get model & user information from Algolia, not server

OAGr:
Previous all models and users were loaded into the browser for each request.  Most of this was redundant, because the important stuff was in the Algolia search results.  

Here I used the Algolia search results for the models and users.  This took a bit of fine tuning, but wasn't too bad. 

This addresses this issue: https://github.com/getguesstimate/guesstimate-app/issues/45


OAGr:
This seems to have improved things, though I also need to add a better server (that's separate functionality though).  Removed model load at 3pm, removed user load at 5pm.

image

PR #106, opened 2016-02-29
Pricing page

OAGr:
This adds a page with pricing information.  This was a bit tricky because it needs different buttons and a possible modal depending on the user state.  In addition, the design required a background color requiring changes to the ``<main>`` component and the style guide.

image

## Style Guide:

image

image


src/components/layouts/main/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+export default class Main extends Component{
+  displayName: 'Main'
+
+  static defaultProps = {
+    isFluid: false
+  }
+
+  render() {
+    const {children, isFluid, backgroundColor} = this.props
```
mmcdermott: PropTypes backgroundColor to be oneof colors we handle atm.

src/components/layouts/main/index.js
```diff
@@ -0,0 +1,33 @@
...
+    if (isFluid) {
+      return (
+        <main className={className}>
+          {children}
+        </main>
+      )
+    } else {
+      return (
+        <main className={className}>
+          <div className='container-fluid wrap'>
```
mmcdermott: This is confusing because 'container-fluid wrap' sounds like a fluid thing to me. Maybe just comment what is going on for now.

src/lib/config/plans.yml
```diff
@@ -6,32 +6,20 @@ plans:
     monthly_cost: 0
     private_model_limit: 0
   -
-    name: 'Tiny'
-    id: 'personal_tiny'
+    name: 'Lite'
+    id: 'personal_lite'
     enum_id: 2
     monthly_cost: 5000
```
mmcdermott: Why milidollars? I think cents would be better.

src/modules/modal/routes.js
```diff
@@ -44,13 +45,15 @@ export default class ModalRouter extends Component{
...
       const isOpen = true
       return(
-        <Modal
-          isOpen={isOpen}
-          onRequestClose={this.props.onClose}
-          style={customStyles}
-        >
-          {<item.component {...props} onClose={this.props.onClose}/>}
-        </Modal>
+          <Modal
```
mmcdermott: Remove extra indentation.

Issue #90, opened 2016-02-17
Toggle privacy of existing models via UI


OAGr:
The 'public' button should be toggle-able.

OAGr:
Addressed by https://github.com/getguesstimate/guesstimate-app/pull/108

PR #108, opened 2016-03-01
Privacy dropdown

OAGr:
This adds a privacy dropdown toggle for spaces:

image

It uses much of the same code (which it refactors) as in the new model privacy toggle:

image

These have alternative states for when a user cannot make new models:

image

image


OAGr:
This addresses the issue https://github.com/getguesstimate/guesstimate-app/issues/90

PR #156, opened 2016-03-07
jStat integration (function based). Closes 154

mmcdermott:


OAGr:
Good work! Surprised it was so few lines.

PR #182, opened 2016-03-14
Fix concat bug 11

mmcdermott:


OAGr:
What is concat bug 11?  Was this a Github issue?

mmcdermott:
Yes, one I just created now. I found it on the plane.

PR #185, opened 2016-03-14
Allowing models to be copied. Closes 115

mmcdermott:
This adds the ability to copy a model. Copying a model creates a new model owned by the copying user, with exactly the same content.

copy-this-model

Currently, copying is only available to signed in users.

### DIFF
--- a/src/components/spaces/show/index.js
+++ b/src/components/spaces/show/index.js
@@ -111,7 +111,9 @@ export default class SpacesShow extends Component {
             <div className='col-sm-10'>
 
               {space &&
-                <SpacesShowHeader onDestroy={this.destroy.bind(this)}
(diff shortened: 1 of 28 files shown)


src/components/spaces/show/index.js
```diff
@@ -106,9 +108,12 @@ export default class SpacesShow extends Component {
             <div className='col-sm-10'>
 
               {space &&
-                <SpacesShowHeader onDestroy={this.destroy.bind(this)}
+                <SpacesShowHeader
+                    isLoggedIn={e.me.isLoggedIn(this.props.me)}
```
OAGr: I'd probably move this logic above the ``return`` method, but it's a very minor issue.

PR #209, opened 2016-03-17
Making lognormal auto-select by default when bounds are positive

mmcdermott:


src/lib/guesstimator/formatter/formatters/DistributionTextUpTo.js
```diff
@@ -1,5 +1,7 @@
 import {normalTextMixin} from './lib.js'
 
+// TODO(matthew): Can we get rid of this????
```
mmcdermott: @OAGr any thoughts on this? Any reason to support this format still?

OAGr: I don't want to get rid of it yet, i personally like the ``:`` syntax and think we may want to change to it eventually.

mmcdermott: Ok, sounds good. May want to revisit this later; not sure we should really support two standards, but not important at the moment. Will remove todo.

src/lib/guesstimator/formatter/formatters/DistributionTextUpTo.js
```diff
@@ -8,9 +8,12 @@ export const item = Object.assign(
     _symbols: ['->', ':'],
     _numbers(text) { return this._splitNumbersAt(text, this._relevantSymbol(text)) },
     format(g) {
-      const guesstimateType = this.guesstimateType(g)
       const [low, high] = this._numbers(g.text)
-      return {guesstimateType, low, high }
+      if (!(isNaN(low) || isNaN(high))) {
```
OAGr: why this line?  This seems like a validation thing, that's dealt with separate.

src/lib/guesstimator/formatter/formatters/lib.js
```diff
@@ -49,7 +49,7 @@ export const normalTextMixin = Object.assign(
         case 'LOGNORMAL':
           return g.guesstimateType
         default:
-          return 'NORMAL'
+          return defaultValue ? defaultValue : ''
```
OAGr: I think ``defaultValue || ''`` would also work.

src/lib/guesstimator/formatter/formatters/lib.js
```diff
@@ -40,7 +40,7 @@ export const normalTextMixin = Object.assign(
   {}, textMixin,
   {
     errors(g) { return this._normalTextErrors(g.text) },
-    guesstimateType(g) {
+    guesstimateType(g, defaultValue) {
```
OAGr: Why not put the line, ``low > 0 ? 'LOGNORMAL' : 'NORMAL'`` here?  This seems like the area where we select the guesstimateType.

mmcdermott: Good idea; I considered that, but incorrectly thought this was imported in more than just the CI selectable things. Will port this over.

src/lib/guesstimator/formatter/formatters/DistributionTextUpTo.js
mmcdermott: The validation code doesn't really seem to stop a lot of the functionality of these pieces right now; invalid input will still pre-define your guesstimate type for example. Will work on making this cleaner, but tbh I think a more general refactor is needed in our validity checking/parsing setup here before this can be really isolated; we basically shouldn't be doing anything with an invalid input.

PR #213, opened 2016-03-17
112 minimal rounding on points and custom data (and calculations thereof)

mmcdermott:
This adds a higher precision setting for point estimates and custom data entries, which persist through calculation. 
varying precision

### DIFF
--- a/src/lib/guesstimator/formatter/formatters/DistributionPointText.js
+++ b/src/lib/guesstimator/formatter/formatters/DistributionPointText.js
@@ -10,7 +10,7 @@ export const item = Object.assign(
     format(g) {
       const {guesstimateType} = this
       const value = parseNumber(g.text)
-      return {guesstimateType, value}
+      return {precision: 6, guesstimateType, value}
     },
     _matchesText(g) { return true },
   }
--- a/src/lib/guesstimator/formatter/formatters/DistributionTextUpTo.js
+++ b/src/lib/guesstimator/formatter/formatters/DistributionTextUpTo.js
@@ -10,7 +10,7 @@ export const item = Object.assign(
     format(g) {
       const guesstimateType = this.guesstimateType(g)
       const [low, high] = this._numbers(g.text)
-      return {guesstimateType, low, high }
+      return {precision: 2, guesstimateType, low, high }
     }
   }
 )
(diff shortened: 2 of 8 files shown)


src/lib/guesstimator/formatter/formatters/DistributionTextUpTo.js
```diff
@@ -10,7 +10,7 @@ export const item = Object.assign(
     format(g) {
       const guesstimateType = this.guesstimateType(g)
       const [low, high] = this._numbers(g.text)
-      return {guesstimateType, low, high }
```
OAGr: I find this quite strange.  If someone estimates, [0.000001,0.00002], would that return as much precision as [5,7]?

src/lib/guesstimator/formatter/formatters/DistributionPointText.js
```diff
@@ -10,7 +10,7 @@ export const item = Object.assign(
     format(g) {
       const {guesstimateType} = this
       const value = parseNumber(g.text)
-      return {guesstimateType, value}
```
OAGr: So if someone enters 0, this would return '0.0000'?

OAGr:
Let's talk about this tomorrow.

src/lib/guesstimator/formatter/formatters/DistributionTextUpTo.js
mmcdermott: Precision is probably poorly named, this just specifies the number of nonzero digits in scientific notation to show for types of estimates. So this actually matches current behavior afaict (e.g. [0.00001, 0.00002] has the same # of nonzero digits as [5,7])

src/lib/guesstimator/formatter/formatters/DistributionPointText.js
mmcdermott: No, it renders as 0. I modified something else that makes it strip trailing zeros from the representation in all cases.

mmcdermott:
Sounds good. This may have been attacking a different issue than intended here; all this PR does is make it so that stats about fixed inputs render with more digits shown in the final representation. It doesn't actually do any real precision/sig-fig propagation.

OAGr:
Let's refactor this to be simpler.

PR #232, opened 2016-03-20
Basic Commerce Tracking (renaming event name to specialty commerce event name).

mmcdermott:
More to come, but this is a minimal, functional bit.


mmcdermott:
To get full advantage out of this, we also need an 'order_id'

OAGr:
Good work

PR #244, opened 2016-03-24
Client interactions for organizations

OAGr:


src/modules/organizations/actions.js
```diff
@@ -0,0 +1,27 @@
+import {actionCreatorsFor} from 'redux-crud'
+import $ from 'jquery' // TODO(matthew): Is this needed at all?
+import * as displayErrorsActions from 'gModules/displayErrors/actions.js'
```
OAGr: if the symbol isn't anywhere in the file (which it doesn't seem to be), it isn't.

src/components/organizations/show/index.js
```diff
@@ -0,0 +1,62 @@
...
+    this.props.dispatch(spaceActions.fetch({organizationId: this.props.organizationId}))
+  }
+
+  render () {
+    const {organizationId, organizations} = this.props
+    const spaces =  _.orderBy(this.props.organizationSpaces.asMutable(), ['updated_at'], ['desc'])
+    // TODO(matthew): Right now, the spaces representer doesn't return either the user or organization (just their
+    // associated ids). As a result, the space list on organization page won't display the user and the space list on
+    // the user page won't display the organizations. Do we want this behavior (in particular for the organizational
+    // case?). If so, should we just embed the organization in the representer, or do a lookup in the spacelistitem.
```
OAGr: let's discuss this

PR #245, opened 2016-03-24
239 organization client interaction space editing

mmcdermott:


src/lib/engine/space.js
```diff
@@ -33,15 +33,25 @@ const user = (space, graph) => {
   return graph.users.find(e => sameIds(e.id, space.user_id))
 }
 
+const organization = (space, graph) => {
+  return graph.organizations.find(e => sameIds(e.id, space.organization_id))
+}
+
+const isMember = (organization_id, user_id, memberships) => (
```
OAGr: why is this in ``space.rb``?

src/lib/engine/space.js
```diff
@@ -33,15 +33,25 @@ const user = (space, graph) => {
...
+  !!_.find(memberships, m => (m.organization_id === organization_id && m.user_id === user_id))
```
OAGr: instead of ``_.find`` you can do ``_.some``

mmcdermott:
So, there's a bug with this, where it basically doesn't try to fetch all the spaces users to display them with the spaces: 

user-missing

My sense is this is lower priority than getting space creation to work, and we shouldn't focus on this for now. I can also just disable showing users at all for this page until we get it to work properly.

src/lib/engine/space.js
```diff
@@ -1,6 +1,7 @@
 import * as _graph from './graph';
 import * as _metric from './metric';
 import * as _guesstimate from './guesstimate';
+import * as UserOrganizationMemberships from 'gEngine/userOrganizationMemberships'
```
OAGr: this should be imported like the stuff above.

src/lib/engine/space.js
```diff
@@ -33,15 +34,21 @@ const user = (space, graph) => {
...
+
 export function toDgraph(spaceId, graph){
   let dGraph = _graph.denormalize(subset(graph, spaceId))
   const space = get(graph.spaces, spaceId)
   const spaceUser = user(space, graph)
+  const userOrganizationMemberships = graph.userOrganizationMemberships
+  const meId = _.get(graph, 'me.id')
   dGraph.user = spaceUser
-  dGraph.ownedByMe = sameIds(_.get(spaceUser, 'id'), _.get(graph, 'me.id'))
+  dGraph.ownedByMe = sameIds(space.user_id, meId) || UserOrganizationMemberships.isMember(space.organization_id, meId, userOrganizationMemberships)
```
OAGr: I would split this into a separate function, with userOrganizationMemberships, which only seems useful for this line.

src/lib/engine/space.js
```diff
@@ -1,6 +1,7 @@
-import * as _graph from './graph';
-import * as _metric from './metric';
-import * as _guesstimate from './guesstimate';
+import * as _graph from './graph'
+import * as _metric from './metric'
+import * as _guesstimate from './guesstimate'
+import * as _UserOrganizationMemberships from './userOrganizationMemberships'
```
OAGr: first character should be lowercase

src/modules/spaces/actions.js
```diff
@@ -57,6 +58,17 @@ export function fetchById(spaceId) {
         const user_id = value.user_id
         const has_user = !!(users.find(e => e.id === user_id))
         if (!has_user) { dispatch(userActions.fetchById(user_id)) }
+
+        // TODO(matthew): Right now, the space has an embedded user and organization record... why are we doing this
```
OAGr: Looking into it a bit, I'm pretty sure that came from before we had the embedded user, so is no longer needed

src/modules/spaces/actions.js
```diff
@@ -57,6 +58,17 @@ export function fetchById(spaceId) {
...
+        // extra fetching?
+        if (value.organization_id) {
+          const organizations = getState().organizations
+          const organization_id = value.organization_id
+          const has_organization = !!(organizations.find(e => e.id === organization_id))
```
OAGr: or ``_.some``

src/modules/spaces/actions.js
mmcdermott: Filed #246 to look into this later.

src/modules/spaces/actions.js
```diff
@@ -57,6 +58,17 @@ export function fetchById(spaceId) {
         const user_id = value.user_id
         const has_user = !!(users.find(e => e.id === user_id))
         if (!has_user) { dispatch(userActions.fetchById(user_id)) }
+
+        // TODO(matthew): Right now, the space has an embedded user and organization record... why are we doing this
+        // extra fetching?
+        if (value.organization_id) {
```
OAGr: I realize that this wasn't done before, but now that this function is so long, I think i would appreciate splitting it up into a helper like ``needs_organization(value.organization_id, getState().organizations)``

src/modules/userOrganizationMemberships/actions.js
```diff
@@ -0,0 +1,33 @@
+import {actionCreatorsFor} from 'redux-crud'
+import $ from 'jquery' // TODO(matthew): Is this needed at all?
```
OAGr: it's not needed unless used in that file

src/modules/userOrganizationMemberships/actions.js
```diff
@@ -0,0 +1,33 @@
...
+import * as displayErrorsActions from 'gModules/displayErrors/actions.js'
+import * as userActions from 'gModules/users/actions.js'
+import {rootUrl} from 'servers/guesstimate-api/constants.js'
+import {captureApiError} from 'lib/errors/index.js'
+import {setupGuesstimateApi} from 'servers/guesstimate-api/constants.js'
+
+let sActions = actionCreatorsFor('userOrganizationMemberships')
+
+function api(state) {
```
OAGr: don't worry about this now, but this definitely needs to get extracted sometime

src/modules/users/actions.js
```diff
@@ -84,3 +84,9 @@ export function create(object) {
   }
 }
 
+export function fetchSuccess(users) {
+  return (dispatch) => {
+    const formatted = users.map(d => _.pick(d, ['auth0_id', 'id', 'name', 'picture']))
```
OAGr: the ``allowedAttributes`` or whatever should be a new function at this point.  It's repeated above.

src/modules/spaces/actions.js
```diff
@@ -52,11 +63,10 @@ export function fetchById(spaceId) {
       }
       else if (value) {
         dispatch(sActions.fetchSuccess([value]))
-
-        const users = getState().users
-        const user_id = value.user_id
-        const has_user = !!(users.find(e => e.id === user_id))
-        if (!has_user) { dispatch(userActions.fetchById(user_id)) }
+        // TODO(matthew): Right now, the space has an embedded user and organization record... why are we doing this
+        // extra fetching?
```
OAGr: it may well be unnecessary now, but we may want to merge this code before refactoring that.

PR #249, opened 2016-03-25
Stop Identifying users before our server's given them a full ID.

mmcdermott:


OAGr:
We really need a commit hook to ensure this gets deployed, right now it takes around 5-10 mins to compile and stuff

PR #263, opened 2016-04-01
Adds Organizations dropdown menu for navigation to organization page. Closes 258.

mmcdermott:
I think this component is rapidly building up some technical debt, but in the interest of speed here's this PR. Note how I am basically ignoring existing mechanisms of drop down closure, as they don't work unless the component renders twice (as on first render this.refs.[name] is undefined). I think we'll need to do a solution like this, or change the way dropdown works (which we should probably do, just the question of now v later).


mmcdermott:
Here's a screenshot of what this PR actually adds:
org-link

PR #273, opened 2016-04-03
Cuts off low density tails of the histogram. Closes 219.

mmcdermott:
This works for heavy tailed distributions, either symmetric or asymmetric. It's deployed to stage now. 

This will not work for anything with significant bimodality.

For uniform & normal distributions over ranges on the order of 10^150 or greater, card histograms don't render but the model remains accessible and usable (semi-graceful failure):
drunken-swan
(though even in these cases the modal still renders the distribution.

For lognormals, things get that bad after ranges on the order of 10^10 or greater:
drunken-swan-log

If those range sizes are unacceptable, we can try to add edge case logic for them somehow.

### DIFF
--- a/src/components/lib/histogram.js
+++ b/src/components/lib/histogram.js
@@ -21,34 +21,42 @@ function avg(arr) {
 }
 
 // Computes min(|a|,|b|)/max(|a|,|b|).
-function fractionLT1(a,b) {
+function fractionLessThanOne(a,b) {
   return Math.min(Math.abs(a),Math.abs(b))/Math.max(Math.abs(a),Math.abs(b))
 }
 
-function filterData(inputData, cutOff) {
+// filterLowDensityPoints removes points that occur in only low density regions of the histogram, to ensure that only
+// points robustly well sampled in the data affect the visualization of the histogram.
+// The parameter 'cutOffRatio' controls how hight te point density must be for points to be kept; a cutOffRatio of 0
+// would keep everything, a cutOffratio of 1.0 would keep nothing.
+// Values that seem to have notable affects are typically > 0.95.
+function filterLowDensityPoints(inputData, cutOffRatio) {
   // We can't filter that intelligently for small sample sets, so we don't bother.
   if (inputData.length < 2000) {
     return inputData
   }
 
-  let outputData = inputData // A copy for immutability
+  let outputData = Object.assign({}, inputData) // A copy for immutability
   outputData.sort((a,b) => a-b) // Sort the data from min -> max.
 
   const bucketSize = outputData.length / 1000 // Grab data in 0.1% chunks.
 
   // Filter Left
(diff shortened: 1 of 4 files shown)


src/components/lib/histogram.js
```diff
@@ -15,12 +15,55 @@ function getXScale(data, width) {
...
+  return Math.min(Math.abs(a),Math.abs(b))/Math.max(Math.abs(a),Math.abs(b))
+}
+
+function filterData(inputData, cutOff) {
+  // We can't filter that intelligently for small sample sets, so we don't bother.
+  if (inputData.length < 2000) {
+    return inputData
+  }
+
+  let outputData = inputData // A copy for immutability
```
OAGr: I'm not sure if this would actually do a deep clone.  The traditional way is to use ``Object.assign({}, inputData)`` or a lodash clone

src/components/lib/histogram.js
```diff
@@ -15,12 +15,55 @@ function getXScale(data, width) {
     nice();
 }
 
+// Computes the average of an array of numbers. If the array is empty, returns 1.
+function avg(arr) {
+  return arr.length > 0 ? arr.reduce((a,b)=>a+b)/arr.length : 1
+}
+
+// Computes min(|a|,|b|)/max(|a|,|b|).
+function fractionLT1(a,b) {
```
OAGr: what does LT1 mean?  Less than 1?  If so, just write that out.

src/components/lib/histogram.js
```diff
@@ -15,12 +15,55 @@ function getXScale(data, width) {
...
+}
+
+function filterData(inputData, cutOff) {
+  // We can't filter that intelligently for small sample sets, so we don't bother.
+  if (inputData.length < 2000) {
+    return inputData
+  }
+
+  let outputData = inputData // A copy for immutability
+  outputData.sort((a,b) => a-b) // Sort the data from min -> max.
```
OAGr: lodash ``sortby`` seems more direct to me.

src/components/lib/histogram.js
```diff
@@ -15,12 +15,55 @@ function getXScale(data, width) {
...
+function avg(arr) {
+  return arr.length > 0 ? arr.reduce((a,b)=>a+b)/arr.length : 1
+}
+
+// Computes min(|a|,|b|)/max(|a|,|b|).
+function fractionLT1(a,b) {
+  return Math.min(Math.abs(a),Math.abs(b))/Math.max(Math.abs(a),Math.abs(b))
+}
+
+function filterData(inputData, cutOff) {
```
OAGr: this name doesn't mean much to me.  Maybe ``intelligentTruncate'' or something more specific, with a description of what exactly it is doing?

src/components/lib/histogram.js
mmcdermott: Leaving as is per offline conversation.

PR #274, opened 2016-04-04
Changed recursive search algorithms to iterative and exit early upon self encounters to prevent infinite loops.

mmcdermott:
This stops the breakage from being as catastrophic, though adds no notice to the user that an infinite loop is the problem (and doesn't communicate that to anywhere else in the program).

### DIFF
--- a/src/lib/basic_graph/basic-graph.js
+++ b/src/lib/basic_graph/basic-graph.js
@@ -13,7 +13,6 @@ export default class BasicGraph {
     return new BasicGraph(itemSubset, edgeSubset)
   }
 
-  // TODO(Question for Ozzie): Why aren't these passing oneLevel down? Seems like this shouldn't work...
   children(id, oneLevel=true) {
     return this.childrenIds(id, oneLevel).map(e => this.nodes.find(m => m.id === e))
   }
--- a/src/lib/basic_graph/basic-graph.js
+++ b/src/lib/basic_graph/basic-graph.js
@@ -15,7 +15,7 @@ export default class BasicGraph {
 
   // TODO(Question for Ozzie): Why aren't these passing oneLevel down? Seems like this shouldn't work...
   children(id, oneLevel=true) {
-    return this.childrenIds(id).map(e => this.nodes.find(m => m.id === e))
+    return this.childrenIds(id, oneLevel).map(e => this.nodes.find(m => m.id === e))
   }
 
   directParents(id) {
--- a/src/lib/basic_graph/basic-graph.js
+++ b/src/lib/basic_graph/basic-graph.js
@@ -13,8 +13,13 @@ export default class BasicGraph {
     return new BasicGraph(itemSubset, edgeSubset)
   }
 
+  // TODO(Question for Ozzie): Why aren't these passing oneLevel down? Seems like this shouldn't work...
   children(id, oneLevel=true) {
-    return this.childrenIds(id, oneLevel).map(e => this.nodes.find(m => m.id === e))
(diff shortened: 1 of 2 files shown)


src/lib/basic_graph/basic-graph.js
```diff
@@ -13,6 +13,7 @@ export default class BasicGraph {
     return new BasicGraph(itemSubset, edgeSubset)
   }
 
+  // TODO(Question for Ozzie): Why aren't these passing oneLevel down? Seems like this shouldn't work...
```
OAGr: you're right, seems like a mistake.  My strong guess is that we don't yet have a case where onelevel is false

src/lib/basic_graph/basic-graph.js
```diff
@@ -22,13 +23,24 @@ export default class BasicGraph {
...
+      }
+      descendants = _.uniq(descendants.concat(newEdges.map(e => e.output)))
+      seen = seen.concat(newEdges)
+      newEdges = this.edges.filter(e => _.some(descendants, d => d.id === e.input))
+    }
+    return descendants
   }
 
+  // TODO(Question for Ozzie): Why didn't I have to fix this too? Is this function ever being called? Is this actually
+  // used?
```
OAGr: I investigated this, it looks like it's not being used now.  We could probably remove ``parents`` and thus ``parentIds``

src/lib/basic_graph/basic-node.js
```diff
@@ -20,11 +20,14 @@ export default class BasicNode {
...
-      return 0
-    } else {
-      const maxChildDistanceFromRoot = Math.max(...this.parents().map(c => c.maxDistanceFromRoot))
-      return (maxChildDistanceFromRoot + 1)
+    // We initialize to -1 as you alwasy get one increment for free in the loop below.
+    let distanceFromRoot = -1
+    let ancestors = [this]
+    while (ancestors.length > 0) {
+      distanceFromRoot++
+      ancestors = _.uniq(_.flatten(ancestors.map(e => e.parents())))
```
OAGr: this could be a bit simpler by doing something like ``_(ancestors).uniq.flatten.filter(e => (e.id !==this.id))``

src/lib/basic_graph/basic-graph.js
```diff
@@ -22,13 +23,24 @@ export default class BasicGraph {
...
-      :
-      _.uniq(_.flattenDeep([oneLevelChildren, oneLevelChildren.map(e => this.childrenIds(e, false))]))
+    let seen = this.edges.filter(e => e.input === id)
+    let descendants = seen.map(e => e.output)
+    if (oneLevel) {return descendants}
+
+    let newEdges = this.edges.filter(e => _.some(descendants, d => d.id === e.input))
+    while (newEdges.length > 0) {
+      if (_.some(newEdges, e => _.some(seen, s => s === e))) {
+        break
```
OAGr: is this basically a breadth first walk?  WIll it stop at the first level with a mistake?

src/lib/basic_graph/basic-graph.js
mmcdermott: AFAICT this isn't being called outside of the library, so I removed it.

src/lib/basic_graph/basic-graph.js
mmcdermott: Yes.

src/lib/basic_graph/basic-graph.js
mmcdermott: It is only used with oneLevel = true, so I'm reducing its scope just to that.

PR #275, opened 2016-04-05
Improving regex and filtering out empty strings after the split to st…

mmcdermott:
…op padding ends of custom data with zeros.


OAGr:
sweet

PR #278, opened 2016-04-06
Distinguish intermediary metrics in UI

OAGr:
Before:
image

After:
image

### DIFF
--- a/src/components/metrics/card/index.js
+++ b/src/components/metrics/card/index.js
@@ -19,11 +19,22 @@ import MetricToken from './token/index.js'
 import './style.css'
 import * as canvasStateProps from 'gModules/canvas_state/prop_type.js'
 
+const INTERMEDIATE = 'INTERMEDIATE'
+const OUTPUT = 'OUTPUT'
+const INPUT = 'INPUT'
+const NOEDGE = 'NOEDGE'
+
+const relationshipClasses = {}
+relationshipClasses[INTERMEDIATE] = 'intermediate'
+relationshipClasses[OUTPUT] = 'output'
+relationshipClasses[INPUT] = 'input'
+relationshipClasses[NOEDGE] = 'noedge'
+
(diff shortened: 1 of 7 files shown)


src/components/metrics/card/index.js
```diff
@@ -158,9 +165,19 @@ class MetricCard extends Component {
 
     const titleView = !this.props.hovered && !isSelected && this._isTitle()
 
+    const relationshipClasses = {
+      INTERMEDIATE: 'intermediate',
```
mmcdermott: I don't think this style of constantitizing the class name actually adds much here; you still have the necessary magic string association between the function above and the map here. why not have the relationshipType function return relationshipClasses.INTERMEDIATE instead of magic strings?

src/lib/engine/dgraph.js
```diff
@@ -23,3 +23,14 @@ export function dependencyMap(dGraph: DGraph): Array<Object>{
   return _.flatten(asLists)
 }
 
+// The bizarro graph is the version of the graph where a guesstimates input is
+// replaces its guesstimte
+export function toBizarroGraph(graph, guesstimateForm){
```
mmcdermott: Comment seems like it is messed up somehow and this name is very noninformative.

src/lib/engine/dgraph.js
```diff
@@ -23,3 +23,14 @@ export function dependencyMap(dGraph: DGraph): Array<Object>{
...
+  //Super slow!
+  let bGraph = _.cloneDeep(graph)
+  bGraph.metrics = bGraph.metrics.map(m => {
+    let guesstimate = (guesstimateForm.metric === m.id) ? guesstimateForm : m.guesstimate
```
mmcdermott: What's gong on here? Why is this sometimes a form and sometimes a guesstimate?

PR #291, opened 2016-04-07
Adding organization_id check into space privacy blocker and doing a b…

mmcdermott:
…it of code style cleanup.

### DIFF
--- a/src/components/spaces/show/index.js
+++ b/src/components/spaces/show/index.js
@@ -70,11 +70,7 @@ export default class SpacesShow extends Component {
   }
 
   onPrivateSelect() {
-    const canMakeMorePrivateModels = e.me.canMakeMorePrivateModels(this.props.me)
-
-    if (canMakeMorePrivateModels) {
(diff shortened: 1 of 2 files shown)


src/components/spaces/show/index.js
```diff
@@ -97,17 +93,17 @@ export default class SpacesShow extends Component {
 
   render() {
     const space = this.props.denormalizedSpace;
-    const sidebarIsViseable = !!space && (space.editableByMe || !_.isEmpty(space.description))
-    const canMakeMorePrivateModels = e.me.canMakeMorePrivateModels(this.props.me)
+    if (!space) { return <div className='spaceShow'></div> }
+
+    const sidebarIsViseable = space.editableByMe || !_.isEmpty(space.description)
+    const canBePrivate = !!space.organization_id || e.me.canMakeMorePrivateModels(this.props.me)
```
OAGr: I feel like this will come back to bite us, if we have free organizations without private functionality.

mmcdermott: Yep, totally agreed (on both it being an issue and on dealing with it then.)

PR #298, opened 2016-04-12
Adds some basic discrete distributions via a package.

mmcdermott:
I don't think these random generation methods are particularly efficient, but they were already existent.

Some examples:
discretes

Additionally, I've aliased the bernoulli distribution with a 'test' function for making multimodal conditional outputs easier:

test

Expanded:

testbig

We may be able to further alias this to make it even easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "karma-webpack": "^1.7.0",
     "mocha": "^2.4.5",
     "packageify": "^0.2.2",
-    "react-addons-perf": "^0.14.7",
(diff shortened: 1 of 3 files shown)


OAGr:
Cool stuff. I'd document this all in our documentation, but remove the discrete distribution, unless we have a good reason for it. (we should have similar functionality in a better other way)

mmcdermott:
Basic documentation added, in line with existent documentation for other distributions.

PR #299, opened 2016-04-12
Add scenarios function

mmcdermott:
This makes an easy construct to add up many values conditioned on individual probabilities. 

Probabilities must sum to 1.
options


mmcdermott:
options2

PR #301, opened 2016-04-19
Transform the ratio taking to ensure that small numbers don't disproportionately affect the results.

mmcdermott:
Some results.
Before (`[1,10]`):
current-1-10
After:
p100-1-10
Before (`[1,1000]`):
current-1-1000
After:
p100-1-1000
Before (`=normal(50,22)`):
current-normal
After:
p100 normal


OAGr:
Tested it out, it seems like a good improvement.

PR #304, opened 2016-04-23
Debouncing functionality for metric card input

OAGr:
We don't debounce the selected metric card, but we do debounce the other results.

Video


OAGr:
http://test1.getguesstimate.com/

PR #309, opened 2016-04-25
Histogram workers

mmcdermott:


OAGr:
I'm going to put this on hold for now.  I don't like the jitter; that the number gets updated at a different time as the graph.

PR #315, opened 2016-04-26
New Range Syntax for Metric Card

OAGr:
## Before
image

## After
image

Not that this doesn't show exactly the 5th to 95th percentiles.  Instead it shows the 10th percentiles for all numbers lower than the mean, and the 90th percentile for all numbers higher than the mean.  This means that it is still a 90% confidence interval, but that now it can not be outside the bounds of the mean.  We've found that when the range is outside the mean some people became very confused, plus the fact that in these cases, the extreme ends would be very significant.

### DIFF
--- a/src/components/distributions/summary/index.js
+++ b/src/components/distributions/summary/index.js
@@ -47,7 +47,7 @@ class DistributionSummarySmall extends Component{
         <div className='mean'>
         <PrecisionNumber value={parseFloat(mean)} precision={precision}/>
         </div>
-        {!!low && !!high && high*low !== 0 &&
+        {!!low && !!high && (low !== high) &&
           <UncertaintyRange low={low} high={high} />
         }
       </div>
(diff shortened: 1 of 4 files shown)


src/components/distributions/summary/index.js
```diff
@@ -31,22 +33,23 @@ class DistributionSummarySmall extends Component{
...
   }
   render () {
-    const {length, mean, stdev, percentiles} = this.props.stats
+    const {length, mean, stdev, adjustedConfidenceInterval} = this.props.stats
 
-    let range = null
-    if (_.isObject(percentiles)) {
-      const [lowRange, highRange] = [(mean - percentiles[5]), (percentiles[95] - mean)]
-      range = (highRange + lowRange) / 2
+    let low
```
mmcdermott: can't you do `let low, high`?

PR #317, opened 2016-04-26
Report infinite loops (and don't show partial loop traversal simulations).

mmcdermott:


src/lib/engine/graph.js
```diff
@@ -46,20 +46,18 @@ export function dependencyList(graph, spaceId) {
...
+  if (metricId) { bGraph = bGraph.subsetFrom(metricId) }
 
-    const nodes =  bGraph.nodes.map(n => [n.id, n.maxDistanceFromRoot])
-    if (notHead){
-      return nodes.filter(e => (e[0] !== metricId))
-    } else {
-      return nodes
-    }
+  const nodes = bGraph.nodes.map(n => [n.id, n.maxDistanceFromRoot])
+  if (onlyHead) {
```
OAGr: The onlyHead check was on top because there are expensive calculations here that it would be nice not to do when it is just the head.

src/lib/propagation/graph-propagation.js
```diff
@@ -36,8 +36,11 @@ export class GraphPropagation {
 
     this.useGuesstimateForm = graphFilters.useGuesstimateForm || false
 
-    this.orderedMetricIds = this._orderedMetricIds(graphFilters)
-    this.orderedMetricPropagations = this.orderedMetricIds.map(id => (new MetricPropagation(id, this.id)))
+    const orderedMetricIdsAndGraphErrors = this._orderedMetricIds(graphFilters)
+    this.orderedMetricIds = orderedMetricIdsAndGraphErrors.map(m => m.id)
+    this.orderedMetricPropagations = orderedMetricIdsAndGraphErrors.map(
```
OAGr: These lines confuse me, let's talk about it briefly.

src/lib/propagation/graph-propagation.js
```diff
@@ -73,9 +76,11 @@ export class GraphPropagation {
...
 
   _orderedMetricIds(graphFilters: object): Array<Object> {
-    if (graphFilters.onlyHead) { return [graphFilters.metricId]}
     this.dependencies = e.graph.dependencyTree(this._graph(), graphFilters)
-    const inOrder = _.sortBy(this.dependencies, function(n){return n[1]}).map(e => e[0])
+    const inOrder = _.sortBy(this.dependencies, function(n){return n[1]}).map(e => ({
+      id: e[0],
+      inInfiniteLoop: !_.isFinite(e[1])
+    }))
     return inOrder
```
OAGr: can you think of a more intuitive way of doing this?

OAGr:
image

PR #336, opened 2016-05-04
Revert 333 revert 330 improve rendering

mmcdermott:
Adds back in component updates, but fixes:
  * Hover state bug
  * Description Text Bug
  * Modal Expansion Bug

should be reviewed, then deployed to stage and smoke tested, then pushed.


OAGr:
Cards shouldn't disappear if one doesn't click enter

PR #337, opened 2016-05-04
Stateful copy paste

mmcdermott:


src/modules/copied/actions.js
```diff
@@ -0,0 +1,46 @@
...
+export function paste(spaceId){
+  return (dispatch, getState) => {
+    const state = getState()
+    if (!(state.copied && state.selection)) { return }
+
+    const {metric, guesstimate} = state.copied
+
+    const location = state.selection
+    const spaceMetrics = getState().metrics.filter(m => m.space === spaceId)
+    const existentMetric = spaceMetrics.find(m => isAtLocation(m, location))
```
OAGr: existingMetric ?

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -45,13 +47,26 @@ export default class FlowGrid extends Component{
 
   state = { rowHeights: [] }
 
-  _handleKeyPress(e) {
+  _handleKeyUp(e){
+    if (e.keyCode == '17') {
+      this.setState({ctrlPressed: false})
```
OAGr: also do for cmd for macs :)

src/modules/copied/actions.js
```diff
@@ -0,0 +1,46 @@
...
+    const existingReadableIds = spaceMetrics.map(m => m.readableId)
+
+    let newMetric = e.metric.create(existingReadableIds)
+    newMetric.location = location
+    const newItem = Object.assign({}, metric, newMetric)
```
OAGr: I'd write these three lines above as, if possible,

``const newItem = {...metric, ...newMetric, location}``

mmcdermott: That doesn't appear to work. Perhaps I'm misunderstanding, but as written that doesn't appear to work.

PR #338, opened 2016-05-04
Show Scatterplot

OAGr:
This adds a new 'View Option' called 'Analysis'.  This option allows you to see the relationship between any two nodes.  Begin by selecting that view option, then select the node in question.  At that point you can see the relationship with each other node, as a scatterplot, complete with its r^2 value.  Upon hover you can see a scatterplot with 1000 samples.

image

## Expanded View
image


src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+class Plot extends Component {
+  render() {
+    const {xSamples, ySamples, size, xLabel, yLabel} = this.props
+    const data = [{ customValues: _.zip(xSamples, ySamples) }];
+    const valuesAccessor = (s) => s.customValues
+    const xAccessor = (e) => e[0]
+    const yAccessor = (e) => e[1]
+
+    if (size === 'SMALL'){ return(
+          <ScatterPlot
```
mmcdermott: It feels like this would be cleaner if you had a function that went from size -> size dependent params and only one return statement. 

e.g. 

function displayParamsForSize(size) {
  if (size === 'SMALL') {
    return {width: 180, height: 68, margin: {top: 5, bottom: 9, left: 5, right: 3}, xTickArgument: 0, xTickSize: 1, ... }
  } else {
     ...
   }
}

... 
const {width, height, margin, ... } = displayParamsFrom(size)
return (<ScatterPlot data={data} width={width} ... />)
...

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+        values={valuesAccessor}/>
+    )}
+  }
+}
+
+class RegressionStats extends Component {
+  render() {
+    const {xSamples, ySamples, size} = this.props
+    const regression = !!xSamples.length ? everpolate.linearRegression(xSamples, ySamples) : false
+    const sampleSize = xSamples && xSamples.length
```
mmcdermott: I wouldn't call this sampleSize b/c you use size for something very different. Maybe sampleCount? or numSamples?

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+        x={xAccessor}
+        y={yAccessor}
+        values={valuesAccessor}/>
+    )}
+  }
+}
+
+class RegressionStats extends Component {
+  render() {
+    const {xSamples, ySamples, size} = this.props
```
mmcdermott: It is strange to me how you have the defensive coding checks for xSamples and not ySamples. What happens if ySamples are undefined? 

Also, can you just pull all your checks to the top? e.g.
if (!xSamples && !xSamples.length && !ySamples && !ySamples.length) {
  return <div></div>
}
...
?

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+export default class SensitivitySection extends Component {
+  static defaultProps = {
+    size: 'SMALL'
+  }
+
+  render() {
+    const {xMetric, yMetric, size} = this.props
+
+    let sampleSize = (size === 'SMALL') ? 100 : 1000
+    const xSamples = _.get(xMetric, 'simulation.sample.values').slice(0, sampleSize)
```
mmcdermott: Does slice even work if xMetric doesn't have a simulation or sample or values? Seems like this would certainly break if xMetric doesn't have simulation.sample.values, in which case I wonder why not just forget the _.get?

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+
+export default class SensitivitySection extends Component {
+  static defaultProps = {
+    size: 'SMALL'
+  }
+
+  render() {
+    const {xMetric, yMetric, size} = this.props
+
+    let sampleSize = (size === 'SMALL') ? 100 : 1000
```
mmcdermott: Ditto prior comment about name.

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
+import React, {Component, PropTypes} from 'react'
+import {BarChart, ScatterPlot} from 'react-d3-components'
+import {sampleMean, sampleStdev, percentile, cutoff, sortDescending} from 'lib/dataAnalysis.js'
```
mmcdermott: you don't appear to use any of these.

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
+import React, {Component, PropTypes} from 'react'
+import {BarChart, ScatterPlot} from 'react-d3-components'
```
mmcdermott: You don't appear to use BarChart here.

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
+import React, {Component, PropTypes} from 'react'
```
mmcdermott: You don't appear to use PropTypes in this file. Maybe remove the import?

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+              <span className='value'> {sampleSize}</span>
+            </div>
+          </div>
+        }
+      </div>
+    )
+  }
+}
+
+export default class SensitivitySection extends Component {
```
mmcdermott: Can we not use export defaults? I find it makes it a fair bit more difficult to trace code paths if you use export defaults

src/components/metrics/card/index.js
```diff
@@ -34,6 +36,16 @@ const relationshipType = (edges) => {
   return NOEDGE
 }
 
+class ScatterTip extends Component {
```
mmcdermott: Seems like this could be a functional component instead of a class based one.

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+import everpolate from 'everpolate'
+import './style.css'
+
+function importance(r2) {
+  if (r2 < 0.05) { return 'low' }
+  else if (r2 < 0.5) { return 'medium' }
+  else { return 'high' }
+}
+
+class Plot extends Component {
```
mmcdermott: Seems like this could be a functional component instead of a class based one.

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+        xAxis={{tickArguments: [6], innerTickSize: 5, outerTickSize: 2, tickPadding: 3, label: xLabel}}
+        yAxis={{tickArguments: [6], innerTickSize: 5, outerTickSize: 2, tickPadding: 3, label: yLabel}}
+        x={xAccessor}
+        y={yAccessor}
+        values={valuesAccessor}/>
+    )}
+  }
+}
+
+class RegressionStats extends Component {
```
mmcdermott: Seems like this could be a functional component instead of a class based one.

src/components/metrics/card/index.js
```diff
@@ -178,10 +190,26 @@ class MetricCard extends Component {
     return errors ? errors.filter(e => !!e) : []
   }
 
+  showSimulation() {
+    const stats = _.get(this.props, 'metric.simulation.stats')
+    if (stats && _.isFinite(stats.mean) && _.isFinite(stats.stdev) && _.isFinite(stats.length)) {
+      return (stats.stdev === 0 || (stats.length > 5))
```
mmcdermott: I think you may want 

`return stats.stdev !== 0`

I'm not sure the length check is relevant and I don't know why you'd want to return true if `stdev === 0` as you have now.

src/components/metrics/card/index.js
```diff
@@ -178,10 +190,26 @@ class MetricCard extends Component {
...
+    } else {
+      return false
+    }
+  }
+
+  _shouldShowSensitivitySection() {
+    const isAnalysis = (this.props.canvasState.metricCardView === 'analysis')
+    return isAnalysis && this.showSimulation() && this.props.selectedMetric
```
mmcdermott: Does this return a boolean or an object? Maybe make `this.props.selectedMetric` `!!this.props.selectedMetric`?

src/components/metrics/card/index.js
```diff
@@ -178,10 +190,26 @@ class MetricCard extends Component {
     return errors ? errors.filter(e => !!e) : []
   }
 
+  showSimulation() {
+    const stats = _.get(this.props, 'metric.simulation.stats')
+    if (stats && _.isFinite(stats.mean) && _.isFinite(stats.stdev) && _.isFinite(stats.length)) {
```
mmcdermott: Why do you check the mean and the length here as well? Could this whole function be simplified to:
```
const stdev = _.get(this.props, 'metric.simulation.stats.stdev')
return !!stdev && stdev !== 0
```
?

src/components/metrics/card/SensitivitySection/SensitivitySection.js
OAGr: We do a few steps in the render method

src/components/metrics/card/updated.js
```diff
@@ -1,5 +1,6 @@
 export function hasMetricUpdated(oldProps, newProps) {
   return (
+    _.get(newProps,'canvasState.metricCardView') === 'analysis' ||
```
mmcdermott: I feel like we could refine this pretty quickly to be more performant. Actually, could this just be: `oldProps.selectedMetric !== newProps.selectedMetric`?

src/components/spaces/canvas/index.js
```diff
@@ -86,15 +85,31 @@ export default class CanvasSpace extends Component{
     this.props.dispatch(changeSelect(next))
   }
 
-  renderMetric(metric) {
+  _selectedMetric() {
+   const {selected} = this.props
+   const metrics = _.get(this.props.denormalizedSpace, 'metrics')
+
+   const item = metrics && metrics.filter(i => _.isEqual(i.location, selected))[0];
```
mmcdermott: Don't need to name item, right? Can just return it directly?

And do we need a defensive check here? Can we push that up higher in the space component, so it short circuits earlier if there are no metrics?

src/components/spaces/canvas/index.js
```diff
@@ -86,15 +85,31 @@ export default class CanvasSpace extends Component{
...
+   return item
+  }
+
+  _isAnalysisView(props = this.props) {
```
mmcdermott: Why does this take a parameter? Just reference this.props inside the function body.

src/components/spaces/canvas/index.js
```diff
@@ -86,15 +85,31 @@ export default class CanvasSpace extends Component{
...
+    return (_.get(props, 'canvasState.metricCardView') === 'analysis')
+  }
+
+  renderMetric(metric, selected) {
     const {location} = metric
+    const hasSelected = selected && metric && (selected.id !== metric.id)
+    const selectedSamples = _.get(selected, 'simulation.sample.values')
+    const passSelected = hasSelected && selectedSamples && !!selectedSamples.length
```
mmcdermott: Here and above, just use selectedSamples.length > 0

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -69,8 +69,8 @@ export default class FlowGrid extends Component{
   }
 
   _columnCount() {
-    const lowestItem = Math.max(...this.props.items.map(e => parseInt(e.location.column))) + 3
-    const selected = parseInt(this.props.selected.column) + 3
+    const lowestItem = Math.max(...this.props.items.map(e => parseInt(e.location.column))) + 4
+    const selected = parseInt(this.props.selected.column) + 4
     return Math.max(6, lowestItem, selected) || 6;
```
mmcdermott: Might be a bit cleaner to put the `+ 4`s in the return. e.g.

return Math.max(2, lowestItem, selected) + 4

Also, I don't think you need the || 6. And if you do you could put it in the map which would be better as then a single `NaN` wouldn't implode the whole calculation (potentially giving too few columns) (e.g. `...(e => parseInt(e.location.cloumn) || 0`)

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+    const xIntercept = regression.intercept && (-1 * (regression.intercept / regression.slope))
+    return (
+      <div className='regression'>
+        {size === 'SMALL' && _.isFinite(rSquared) &&
+          <div>
+            <span className='label'> r<sup>2</sup></span>
+            <span className={`value ${importance(rSquared)}`}> {rSquared.toFixed(2)}</span>
+          </div>
+        }
+        { size !== 'SMALL' &&
```
mmcdermott: why no `_.isFinite(rSquared)` check here?

src/components/metrics/card/SensitivitySection/SensitivitySection.js
```diff
@@ -0,0 +1,113 @@
...
+          <div>
+            <div>
+              <span className='label'> r<sup>2</sup></span>
+              <span className='value'> {regression.rSquared.toFixed(2)}</span>
```
mmcdermott: nit: odd formatting, here and below, with space after `>` but not before `<`

src/modules/canvas_state/prop_type.js
```diff
@@ -10,6 +10,7 @@ export const metricCardView = PT.oneOf([
   'normal',
   'basic',
```
mmcdermott: I thought we removed basic mode?

src/modules/canvas_state/prop_type.js
```diff
@@ -10,6 +10,7 @@ export const metricCardView = PT.oneOf([
...
   'scientific',
+  'analysis',
   'display',
   'debugging'
```
mmcdermott: Also debugging is removed, yes?

PR #341, opened 2016-05-05
Don't mutate the samples.

mmcdermott:


mmcdermott:
At least, this makes the bug harder to spot. I don't expect this eliminates the bug. Still an improvement though, while I find the real culprit. Will update when I've verified that this does or does not fix the true underlying issue.

mmcdermott:
I lied; this does appear to completely solve the issue. I'm not sure why it stops the truncation (as opposed to just the ordered truncation), but it certainly appears to.

PR #345, opened 2016-05-09
UI improvements

OAGr:
- Fixes a bug where selecting on a distribution shape would select the next cell (selection would happen on mouseUp, while the click happened on mouseDown).
- Metric is deselected after being opened. This stops a bug regarding input.
- The metric input no longer gets squished for mobile devices.  However, there still seems to be a bug for large empty cells.
- The Guesstimate form is a bit more small, to go along with recent changes to the cards.
- The value text box now expands as one types, similar to the name field.

### DIFF
--- a/src/components/distributions/editor/TextForm/TextForm.js
+++ b/src/components/distributions/editor/TextForm/TextForm.js
@@ -57,34 +57,30 @@ export default class TextForm extends Component{
     const {input} = guesstimateForm
     const guesstimateType = this._guesstimateType()
     return(
-      <div>
-        <div className='row'>
-          <div className='col-sm-12'>
-            <TextInput
-              value={input}
-              onEscape={onEscape}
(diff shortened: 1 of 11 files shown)


src/components/distributions/editor/TextForm/TextForm.js
```diff
@@ -95,26 +103,22 @@ export default class TextForm extends Component{
...
-            </div>
-            <div className='col-sm-4'>
-              {hasEmptyInput &&
-                <a className='custom-data' onClick={this.props.onAddDefaultData}>
-                  Add Custom Data
-                </a>
-              }
-            </div>
+    if (!isLarge) { return( this._textInput() ) }
+    else {
```
mmcdermott: don't need the else here as these are return blocks.

OAGr: ignoring for now

PR #346, opened 2016-05-09
Don't resample data and show data histograms for even small data sets.

mmcdermott:
The two datasets below are negated versions of each other, so their sum should be 0 in every sample.
no-resample-inputs


mmcdermott:
Uploading to stage.getguesstimate.com now; be sure to try and break it before merging it (I have not been able to).

mmcdermott:
I spoke too soon; opening modal reorders custom data still. Will investigate and re-open.

PR #371, opened 2016-05-14
Multiple select background experiment

OAGr:
image

### DIFF
--- a/src/components/lib/FlowGrid/FlowGrid.jsx
+++ b/src/components/lib/FlowGrid/FlowGrid.jsx
@@ -12,8 +12,6 @@ import {isWithinRegion, isAtLocation} from 'lib/locationUtils.js'
 import HTML5Backend from 'react-dnd-html5-backend';
 import { DragDropContext } from 'react-dnd';
 
-import GridPoint from './gridPoints.js'
-
 let upto = (n) => Array.apply(null, {length: n}).map(Number.call, Number)
 
 const PTLocation = PropTypes.shape({
--- a/src/components/lib/FlowGrid/FlowGrid.jsx
+++ b/src/components/lib/FlowGrid/FlowGrid.jsx
@@ -129,6 +129,21 @@ export default class FlowGrid extends Component{
     return Math.max(6, lowestItem, selected) || 6;
   }
 
+  _selectionRelation(location) {
+    const {multipleSelected} = this.props
+    let relation = {within: false, onLeft: false, onRight: false, onTop: false, onBottom: false}
+
+    relation.within = (multipleSelected[0].row <= location.row && multipleSelected[0].column <= location.column &&
+       multipleSelected[1].row >= location.row && multipleSelected[1].column >= location.column)
+
+    relation.onTop = multipleSelected[0].row === location.row
+    relation.onBottom = multipleSelected[1].row === location.row
+    relation.onLeft = multipleSelected[0].column === location.column
+    relation.onRight = multipleSelected[1].column === location.column
+
+    return relation
+  }
+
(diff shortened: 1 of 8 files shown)


src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -32,7 +34,7 @@ export default class FlowGrid extends Component{
       output: PTLocation.isRequired
     })),
     selected: PTLocation,
-    multipleSelected: PropTypes.arrayOf(PTLocation, PTLocation),
+    showRegion: PropTypes.arrayOf(PTLocation, PTLocation),
```
mmcdermott: Per conversation, rename to selectedRegion

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -129,16 +131,21 @@ export default class FlowGrid extends Component{
     return Math.max(6, lowestItem, selected) || 6;
   }
 
+  _isWithin(location) {
```
mmcdermott: We have this function (or an equivalent) in many places; in a separate PR I'm making a location utils file with it and a few others, so eventually I'll strip this function out. Just FYI.

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -3,14 +3,16 @@ import React, {Component, PropTypes} from 'react'
...
-import EdgeContainer from './edge-container.js'
+import BackgroundContainer from './background-container.js'
 
 import {keycodeToDirection, DirectionToLocation} from './utils'
 import {removeMetric} from 'gModules/metrics/actions.js'
 
 import HTML5Backend from 'react-dnd-html5-backend';
 import { DragDropContext } from 'react-dnd';
 
+import GridPoint from './gridPoints.js'
```
mmcdermott: Why do you need this import here?

src/components/lib/FlowGrid/background-container.js
```diff
@@ -38,13 +48,18 @@ export default class EdgeContainer extends Component {
...
+    if (!columnWidth || !rowHeights.length){ return false }
     return (
-      <Edges
-          columnWidth={columnWidth}
-          containerHeight={containerHeight}
-          edges={edges}
-          rowHeights={rowHeights}
-      />
+      <div>
+        <Edges
```
mmcdermott: We should note here that we will rename 'Edges' -> 'Arrows' with a comment, so I can find it easily later in a refactor. Shouldn't make that change now, though.

src/components/lib/FlowGrid/background-container.js
```diff
@@ -2,13 +2,22 @@ import React, {Component, PropTypes} from 'react'
 import Dimensions from 'gComponents/utility/react-dimensions';
 import Edges from './edges';
 import $ from 'jquery'
-import _ from 'lodash'
+import GridPoint from './gridPoints.js'
+
+export class Region extends Component{
```
mmcdermott: This is never imported, right? So we can not export it?

src/components/lib/FlowGrid/gridPoints.js
```diff
@@ -0,0 +1,40 @@
+let upto = (n) => Array.apply(null, {length: n}).map(Number.call, Number)
+
+export default class GridPoint {
```
mmcdermott: This doesn't warrant a change right now, but I would say that making this a functor as opposed to just having a function in this library that returns another function seems a bit strange if we aren't passing it as props or otherwise checking for equality somewhere.

PR #373, opened 2016-05-14
Multiple select

mmcdermott:


src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -56,56 +58,87 @@ export default class FlowGrid extends Component{
...
+
     let direction = keycodeToDirection(e.keyCode)
     if (direction) {
       e.preventDefault()
       const size = ({columns: this._columnCount(), rows: this._rowCount()})
-      let newLocation = new DirectionToLocation(size, this.props.selected)[direction]()
+      let newLocation = new DirectionToLocation(size, this.props.selectedCell)[direction]()
       this.props.onSelectItem(newLocation)
-    } else if (e.keyCode == '17' || e.keyCode == '224' || e.keyCode == '91') {
+      this.refs[`cell-${newLocation.row}-${newLocation.column}`].decoratedComponentInstance._focus()
```
OAGr: is this still necessary?

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -56,56 +58,87 @@ export default class FlowGrid extends Component{
...
+  _handleEndRangeSelect(corner1) {
+    const corner2 = this.props.selectedCell
+
+    if (!isLocation(corner2)) {
+      this.props.onSelectItem(corner1)
+      return
+    }
+
+    this.props.onMultipleSelect(corner1, corner2)
+  }
```
OAGr: what if corner1 == corner2? Also, any chance this above function could be cleaned up a bit more?

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -132,6 +165,10 @@ export default class FlowGrid extends Component{
     }
   }
 
+  componentWillUnmount() {
+    this.props.onDeSelectAll()
+  }
```
OAGr: I think this should go in the space canvas, not here.

src/components/lib/FlowGrid/FlowGrid.jsx
mmcdermott: Multi-selecting a 1x1 grid is perfectly valid; actually is used often. Right now I believe region select controls most of the styling, so should be present on singleton selections as well (and is presently). I'll add an else statement momentarily.

src/components/lib/FlowGrid/FlowGrid.jsx
mmcdermott: This was a bit of future proofing, I guess/API separation. The space canvas also de-selects all upon unmount, but I think the FlowGrid should too. If you were to use this component, and it were to unmount, then the flowgrid has disappeared from the page; I don't think in that case it is sensible to retain which flowgrid locations were selected.

src/components/spaces/canvas/index.js
```diff
@@ -76,15 +78,21 @@ export default class Canvas extends Component{
...
+    this._handleDeSelectAll()
   }
 
   _handleSelect(location) {
     this.props.dispatch(changeSelect(location))
+    this.props.dispatch(selectRegion(location, location))
   }
 
-  _handleDeSelect() {
+  _handleMultipleSelect(corner1,corner2) {
```
OAGr: space in between var names

src/lib/propagation/graph-propagation.js
```diff
@@ -39,7 +40,7 @@ export class GraphPropagation {
     const orderedMetricIdsAndGraphErrors = this._orderedMetricIdsAndErrors(graphFilters)
     this.orderedMetricIds = orderedMetricIdsAndGraphErrors.map(m => m.id)
     this.orderedMetricPropagations = orderedMetricIdsAndGraphErrors.map(
-      ({id, errors}) => (new MetricPropagation(id, errors, this.id))
+      ({id, errors}) => (new MetricPropagation(id, errors, this.id, !!this.onlyMetrics && !_.some(this.onlyMetrics, m => m.id === id)))
```
OAGr: split into two lines if possible.

src/modules/copied/actions.js
```diff
@@ -1,43 +1,74 @@
...
 
-    const newGuesstimate = Object.assign({}, guesstimate, {metric: newItem.id})
+    const existingMetrics = spaceMetrics.filter(m => isWithinRegion(m.location, pasteRegion))
+    existingMetrics.forEach(existingMetric => {dispatch(metricActions.removeMetric(existingMetric.id))})
 
-    const existingMetric = spaceMetrics.find(m => isAtLocation(m, location))
-    if (existingMetric) {
-      dispatch(metricActions.removeMetric(existingMetric.id))
-    }
+    newMetrics.forEach((newMetric, i) => {dispatch({type: 'ADD_METRIC', item: newMetric, newGuesstimate: newGuesstimates[i]})})
```
OAGr: would be nice for this to be shorter

mmcdermott: Agreed, going to leave off for now though as I don't think there is significant functionality we can extract here that wouldn't be just extracting function scope and putting it higher in the file, which IMO would increase total cost of ownership.

OAGr:
Woo! Good job

PR #375, opened 2016-05-14
Revert "When card is dragged, do not show background"

OAGr:
Reverts getguesstimate/guesstimate-app#372


OAGr:
This caused problems with focus.

PR #380, opened 2016-05-16
Minor styling changes to green button, space user icon, header on mobile

OAGr:
## Before
image
image

## After

image

image


src/routes/layouts/login/index.js
```diff
@@ -182,15 +182,11 @@ export default class Profile extends Component {
       { isLoggedIn && this.profileDropdown() }
 
       { !isLoggedIn &&
-        <a className={'item text'} href='/pricing'>Plans</a>
-      }
-
-      { !isLoggedIn &&
-        <a className={'item text'} href='/models'>Explore</a>
```
mmcdermott: Why are you getting rid of the header link to the public models page? Or am I misreading this?

mmcdermott:
I kind of liked the green, actually, but I trust your design sense over mine. Other than my question about the 'Explore' link on the header, this lgtm. If that deletion was intentional, then feel free to merge and close, and we can talk about it later, I just wasn't sure as it wasn't mentioned as a change in your description.

OAGr:
I designed with Flux on, so the blue wasn't as apparent, definitely agree the green you saw wasn't great.  Changed it.

The deletion of the 'explore' word was intentional, kind of part of the 'mobile' push.

mmcdermott:
I'd be interested in exploring ways to add back in 'explore' down the road as that is where a lot of our users spend their time right now, and is basically our only portal to the 'browse' behavior we may want to enable.

I'm going to wait on merging this a bit; the metric readable IDs seem very close to the right in your picture; it looked like something was a bit off to me. May pull it and play with it to get a better feel.

PR #382, opened 2016-05-16
A hacky, definitive way to exclude ourselves from data analytics, regardless of provider.

mmcdermott:


mmcdermott:
Not ready yet; only did identify.

PR #388, opened 2016-05-17
Some Miscellaneous refactors, stopping some unnecessary requests, and narrowing the props passed to space subcomponents.

mmcdermott:


src/modules/spaces/actions.js
```diff
@@ -67,11 +66,18 @@ export function fetchById(spaceId) {
         captureApiError('SpacesFetch', null, null, err, {url: 'spacesfetch'})
       }
       else if (value) {
+        debugger
```
OAGr: debugger

PR #393, opened 2016-05-18
Apply Seamless Immutability to our reducers in dev mode so we don't allow state changing bugs to happen.

mmcdermott:


src/modules/reducers.js
```diff
@@ -1,4 +1,6 @@
 import reduxCrud from 'redux-crud'
+import SeamlessImuutable from 'seamless-immutable'
+const SI = __DEV__? SeamlessImuutable : (a) => a
```
OAGr: I find this pretty funny for some reason.

OAGr:
This is a clean chance, nice.  Curious how slow this will make local development, but we'll find that out.

PR #404, opened 2016-05-21
Main page update

OAGr:
A new card layout for spaces.  This is possible and more reasonable now that we have more space images. This impacts the main search page and the user pages, but does not yet impact that organizational pages. 

# Before
image

image

# After 

image

image


src/components/spaces/cards/index.js
```diff
@@ -0,0 +1,83 @@
+import React from 'react'
```
mmcdermott: Can you do the imports for consistency with how I've been ordering them? Like this:
```
import React from 'react'

import Icon from 'react-fa'

import * as navigationActions from 'gModules/navigation/actions'

import * as Space from 'gEngine/space'
import * as User from 'gEngine/user'

import {formatDescription, formatDate} from 'gComponents/spaces/shared'

import arrowsVisibleImage from '../../../assets/metric-icons/blue/arrows-visible.png'

import './style.css'
```
Ordered by sections like this, new lines separating sections:
    - React
    - Redux
    - NPM modules
    - Internal components
    - Internal selectors
    - Internal actions
    - Engine
    - Internal libs
    - servers
    - Images
    - Stylesheets

src/components/spaces/cards/index.js
```diff
@@ -0,0 +1,83 @@
...
+
+import arrowsVisibleImage from '../../../assets/metric-icons/blue/arrows-visible.png'
+import {formatDescription, formatDate} from 'gComponents/spaces/shared'
+import './style.css'
+
+import * as navigationActions from 'gModules/navigation/actions.js'
+import * as Space from 'gEngine/space';
+import * as User from 'gEngine/user';
+
+let BlankScreenshot = () => (
```
mmcdermott: const

src/components/spaces/cards/index.js
```diff
@@ -0,0 +1,83 @@
...
+let SingleButton = ({isPrivate}) => (
+  <div className='tag'>
+    <Icon name={isPrivate ? 'lock' : 'globe'}/>
+  </div>
+)
+
+let ButtonArea = ({user, navigateToUser, isPrivate, showPrivacy}) => (
+  <div className='hover-row'>
+    {user &&
+    <div className='user-tag' onClick={navigateToUser}>
```
mmcdermott: Indent within the conditional block?

src/components/spaces/cards/index.js
```diff
@@ -0,0 +1,83 @@
...
+      <img
+          className='avatar'
```
mmcdermott: 2 spaces

src/components/spaces/cards/index.js
```diff
@@ -0,0 +1,83 @@
...
+          src={user.picture}
+      />
+      <div className='name'>
+      {user.name}
```
mmcdermott: indent

src/components/spaces/cards/index.js
```diff
@@ -0,0 +1,83 @@
...
+    </div>
+  )
+}
+
+const SpaceCards = ({spaces, showPrivacy}) => (
+  <div className='row'>
+    {_.map(spaces, (s) =>
+        <SpaceCard
+          key={s.id}
+          space={s}
```
mmcdermott: Can we narrow these props here a bit (e.g. pass space.name, space.id, space.user, and space.description directly)? I don't like these depending on the graphs. This probably isn't a big deal, as the graphs aren't actually being touched on this page, but nevertheless, might be a good idea. Your call, as I don't think this will really have a big impact.

src/components/spaces/index/index.js
```diff
@@ -1,21 +1,54 @@
 import React, {Component, PropTypes} from 'react'
```
mmcdermott:
```
import React, {Component} from 'react'
import {connect} from 'react-redux'

import Icon from'react-fa'

import SpaceCards from 'gComponents/spaces/cards'

import * as search from 'gModules/search_spaces/actions'

import './style.css'
```

src/components/spaces/list_item/index.js
```diff
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import { denormalizedSpaceSelector } from '../denormalized-space-selector.js';
```
mmcdermott: This file definitely needs a refactor, but I'll leave that for a separate PR.

OAGr: I figure we'll eventually just remove the list_item, as that was the old way of doing it (now only being used by companies)

src/components/spaces/cards/index.js
OAGr: I thought about this a bit, think we can do it later.

src/components/spaces/list_item/index.js
mmcdermott: SG, makes sense.

PR #416, opened 2016-05-24
Added background color for copied region

OAGr:
The intended functionality is that the copied region of a canvas appears to have a dashed border around it.  After one paste, this goes away.  

One downside is that it's impossible to remove this selection without pasting, but I believe this is how similar tools do it.

image

### DIFF
--- a/src/modules/copied/reducer.js
+++ b/src/modules/copied/reducer.js
@@ -1,7 +1,9 @@
 export const copiedR = (state = null, action) => {
   switch (action.type) {
     case 'COPY':
-      return action.copied
+      return {...action.copied, pastedTimes: 0}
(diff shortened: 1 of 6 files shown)


src/modules/copied/reducer.js
```diff
@@ -1,7 +1,9 @@
 export const copiedR = (state = null, action) => {
   switch (action.type) {
     case 'COPY':
-      return action.copied
+      return {...action.copied, pastedTimes: 0}
+    case 'COPY_PASTE':
```
mmcdermott: Maybe just PASTE? COPY_PASTE seems odd to me. You're just pasting, right?

PR #425, opened 2016-05-26
Migrate metric card names and value fields to use draft.js

OAGr:
This in ways simplifies the codebase, will remove various bugs around the text fields, and will allow for much more functionality and control around the text areas.

Bug fixes/improvements
- Text areas should be relatively bug-free. There shouldn't be hidden or scrolling issues.
- Cursor becomes pointer when in metricSelection state.
- Clicking backspace will typically not go to browser history.
- After once cell is selected, it should only take one click to select an empty cell.  It used to take 2.
- Undo / redo should completely work when outside of text areas, and should influence the text when in text areas.
- When metric card name is edited, background area should not get out of sync (forceUpdate)
- Clicking delete on empty cell shouldn't cause an error in the console.
- Refactored GuesstimateForm so that it's a prop to the space canvas, and only passed into the relevant metric card.  It should only display if has correct metric ID.


src/components/lib/FlowGrid/cell.jsx
```diff
@@ -76,24 +76,29 @@ export default class Cell extends Component {
...
-      return
+    const isFunctionSelect = (this.props.canvasState.metricClickMode === 'FUNCTION_INPUT_SELECT')
+    const {inSelectedCell, item, location} = this.props
+    const leftClick = (e.button === 0)
+
+    if (!leftClick) { return }
+
+    if (!item){ this.props.onEmptyCellMouseDown(e) }
+
+    if (inSelectedCell && !item){
```
OAGr: space before curly brace

src/components/metrics/card/MetricCardViewSection/style.css
```diff
@@ -66,6 +65,14 @@
   flex-direction: column;
 }
 
+.MetricCardViewSection.anotherFunctionSelected > .StatsSection{
```
OAGr: space before curly

src/components/metrics/card/index.js
```diff
@@ -265,10 +266,4 @@ class MetricCard extends Component {
...
 }
 
-function select(state) {
-  return {
-    guesstimateForm: state.guesstimateForm
-  }
-}
-
-module.exports = connect(select)(MetricCard);
+module.exports = connect()(MetricCard);
```
OAGr: does this still need the connect?

src/components/distributions/editor/TextForm/TextInput.js
```diff
@@ -6,7 +6,57 @@ import TextArea from 'react-textarea-autosize'
 
 import DistributionSelector from './DistributionSelector'
 
-import insertAtCaret from 'lib/jquery/insertAtCaret'
+import {EditorState, Editor, ContentState, getDefaultKeyBinding, KeyBindingUtil, Modifier} from 'draft-js'
```
mmcdermott: Put this import up in the section with the other npm modules (just below 'TextArea').
(which, I'm guessing isn't needed anymore, as well? Also, ReactDOM, it appears?)

src/components/distributions/editor/TextForm/TextInput.js
```diff
@@ -6,7 +6,57 @@ import TextArea from 'react-textarea-autosize'
...
+
+class SimpleEditor extends React.Component {
```
mmcdermott: just 'Component', right?

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -99,7 +99,8 @@ export default class FlowGrid extends Component{
   }
 
   _handleRemoveSelectedItems() {
-    this.props.onRemoveItems(this._selectedItems().map(i => i.key))
+    const selectedItems = this._selectedItems()
+    !_.isEmpty(selectedItems) && this.props.onRemoveItems(this._selectedItems().map(i => i.key))
```
mmcdermott: This feels like the wrong place for defensivity; onRemoveItems should be able to handle an empty list, I would say.

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -115,7 +116,7 @@ export default class FlowGrid extends Component{
       const size = ({columns: this._columnCount(), rows: this._rowCount()})
       let newLocation = new DirectionToLocation(size, this.props.selectedCell)[direction]()
       this.props.onSelectItem(newLocation)
-    } else if (!e.shiftKey && (e.keyCode == '17' || e.keyCode == '224' || e.keyCode == '91')) {
+    } else if (!e.shiftKey && (e.keyCode == '17' || e.keyCode == '224' || e.keyCode == '91' || e.keyCode == '93')) {
```
mmcdermott: Why are you adding the alt key (93) here?

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -128,6 +129,7 @@ export default class FlowGrid extends Component{
         this._handleRemoveSelectedItems()
       } else if (e.keyCode == '90' && !e.shiftKey) {
         this.props.onUndo()
+        e.preventDefault()
```
mmcdermott: I'm surprised that only undo needs the prevent default... Why not redo?

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -170,8 +172,10 @@ export default class FlowGrid extends Component{
     const item = this.props.items.find(i => isAtLocation(i.location, location));
     return (
       <Cell
-        hasItemUpdated={this.props.hasItemUpdated}
+        canvasState={this.props.canvasState}
+        forceFlowGridUpdate={() => this.forceUpdate()}
```
mmcdermott: Yikes. Why is this necessary?

src/components/lib/FlowGrid/FlowGrid.jsx
OAGr: The mac has multiple cmd keys, 91 and 93.  These are  on both sides of the keyboard.

src/components/metrics/card/index.js
mmcdermott: If you're not connecting it to anything, I doubt it, but I'm not sure.

src/components/lib/FlowGrid/FlowGrid.jsx
OAGr: Happy to explain on phone.

src/components/metrics/card/name/index.js
```diff
@@ -1,108 +1,118 @@
 import React, {Component, PropTypes} from 'react'
 
 import TextArea from 'react-textarea-autosize'
```
mmcdermott: Don't need TextArea anymore, I think.

src/components/metrics/card/name/index.js
```diff
@@ -1,108 +1,117 @@
 import React, {Component, PropTypes} from 'react'
 
-import TextArea from 'react-textarea-autosize'
+import {EditorState, Editor, ContentState, getDefaultKeyBinding, KeyBindingUtil} from 'draft-js'
 
 import './style.css'
 
+class SimpleEditor extends React.Component {
```
mmcdermott: Assuming this is the place you said you changed it but it wasn't showing up.

src/components/metrics/card/name/index.js
```diff
@@ -1,108 +1,117 @@
...
-      return this.state.editing
-    } else {
-      return this.props.editable && this.state.editing
-    }
+    value: this.props.name
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.name === this.state.value) {this.setState({value: nextProps.name})}
+    if (!nextProps.inSelectedCell && (this.props.name !== nextProps.name) && (nextProps.name !== this.value())) {
```
mmcdermott: This change causes a bug with undo. To reproduce, change a metric's name, then select the cell (but not the text field) and hit ctrl-z. The name won't change back (but you can tell by the metric flickering that undo is being called).

src/components/metrics/card/name/index.js
```diff
@@ -1,108 +1,117 @@
...
+  _onChange(editorState) {
+   return this.setState({editorState})
+  }
+
+  focus() {
+    this.refs.editor.focus()
+  }
+
+  changePlainText(value) {
+    this.setState({editorState: this._plainTextEditorState(value)})
```
mmcdermott: you did something pretty similar a different way in the previous SimpleEditor component. Why do them different ways?

src/components/metrics/card/name/index.js
mmcdermott: It'd be nice if we could not name two separate components the same thing. Why not NameEditor?

src/components/metrics/card/name/index.js
```diff
@@ -1,108 +1,117 @@
...
-  onChange(e) {
-    this.setState({value: e.target.value})
+  value() {
+    return this.refs.SimpleEditor.getPlainText()
   }
 
   handleKeyDown(e) {
+    e.stopPropagation()
+    this.props.heightHasChanged()
     const ENTER = (e) => ((e.keyCode === 13) && !e.shiftKey)
```
mmcdermott: This ENTER stuff isn't working AFAICT. Can we just remove it?

src/components/metrics/card/name/index.js
```diff
@@ -1,108 +1,117 @@
...
-        {!this._editable() &&
-          <div className={`static${!this.state.value ? ' default-value' : ''}`}
-            onMouseOver={() => {if (!this.state.editing) {this.setState({editing: true})}}}n
-            onMouseDown={() => {this.setState({persistEditing: true})}}
-          >
-            {(this.state.value || 'name').replace(/ /g, "\u2005")}
-          </div>
-        }
-      </div>
+      <span
```
mmcdermott: Just for my information, why change from div to span?

src/components/utility/NavHelper/index.js
```diff
@@ -18,6 +19,14 @@ export default class NavHelper extends Component{
     }
   }
 
+  componentDidMount(){
```
mmcdermott: Why this change? What is this for?

src/modules/metrics/actions.js
```diff
@@ -22,8 +22,8 @@ export function addMetric(item) {
 
 //spaceId must be done before the metric is removed here.
 export function removeMetrics(ids) {
-  if (ids.length === 0) { return }
   return (dispatch, getState) => {
+    if (ids.length === 0) { return }
```
mmcdermott: Why move the edge case return?

src/components/utility/NavHelper/index.js
OAGr: This makes sure that clicking 'delete' when targeting the body wouldn't go to previous page.  A few users, including myself, experienced an issue where this would cause them to leave a canvas.  Apparently the area they targeted (while on the spaces show page) was the body.

src/components/metrics/card/name/index.js
OAGr: No strong reason, but I generally prefer spans for things that won't take extra space.  They're less flexible, which makes that more explicit imo.

src/modules/metrics/actions.js
OAGr: When it was above, it wouldn't have been a valid action.  That's what caused the error.

src/components/metrics/card/name/index.js
OAGr: I think we should leave it for now, may want to make it work soon instead, and removing all the related code would take a bit.

src/components/metrics/card/name/index.js
OAGr: They're similar but different.  Here we wanted some way that the text could be overridden completely from the parent element.  There we wanted to return the planText on every key stroke.  

We could refactor them to make them more similar or the same thing, but I expected to make them even more different later on.

src/components/metrics/card/name/index.js
mmcdermott: If you're going to keep it at least remove the stopPropagation line (which is unnecessary now) and add a comment that this is currently not working.

Side note, I don't know if we need this at all regardless, as tabbing works fine to jump between sections (and seems more intuitive to me. I would expect Enter to jump out of the cell entirely, perhaps, but not to jump between sections I don't think).

src/modules/metrics/actions.js
mmcdermott: What error?

OAGr: There was a redux error shown in the console when an empty cell was used for deletion

src/components/metrics/card/name/index.js
OAGr: Don't think this is worth another commit here as this PR.

PR #434, opened 2016-05-30
Auto select metric text forms

OAGr:
- For metric name, 'return' key selects value field.  Shift-return goes to a newline.
- For value field, 'return' key selects metric card.  Shift-return goes to a newline.
- Empty selected cards start off with name fields being selected.

Example currently on stage:
http://stage.getguesstimate.com/

### DIFF
--- a/src/components/distributions/editor/TextForm/TextInput.js
+++ b/src/components/distributions/editor/TextForm/TextInput.js
@@ -33,6 +33,16 @@ class TextInputEditor extends Component {
     return this.setState({editorState})
   }
 
+  handleReturn(e) {
+    const shifted = e.shiftKey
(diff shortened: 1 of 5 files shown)


src/components/distributions/editor/TextForm/TextInput.js
```diff
@@ -33,6 +33,16 @@ class TextInputEditor extends Component {
     return this.setState({editorState})
   }
 
+  handleReturn(e) {
+    const shifted = e.shiftKey
```
mmcdermott: I'm not sure naming this 'shifted' adds any meaningful clarity; I'd remove it and just test on e.shiftKey directly.

PR #440, opened 2016-05-31
Probably needs a refactor later, but stops us from saving every key stroke.

mmcdermott:


src/components/distributions/editor/TextForm/TextInput.js
```diff
@@ -28,6 +28,10 @@ class TextInputEditor extends Component {
     return editorState.getCurrentContent().getPlainText('')
   }
 
+  _onBlur() {
+    this.props.onBlur(this.state.editorState.getCurrentContent().getPlainText(''))
+  }
```
OAGr: Why does this need to pass the value to the parent? The parent element, TextInput, already has that value in state.value.

src/components/distributions/editor/TextForm/TextInput.js
```diff
@@ -86,9 +90,9 @@ export default class TextInput extends Component{
     this.props.onFocus()
   }
 
-  _handleBlur() {
+  _handleBlur(value) {
     $(window).off('functionMetricClicked')
-    this.props.onBlur()
+    this.props.onBlur(value)
```
OAGr: could probably just be `this.props.onBlur(this.state.value)`

src/components/distributions/editor/TextForm/TextInput.js
```diff
@@ -73,7 +77,7 @@ export default class TextInput extends Component{
     value: PropTypes.string,
   }
 
-  componentWillUnmount() { this._handleBlur() }
+  componentWillUnmount() { this._handleBlur(this.props.value) }
```
OAGr: Is this needed? I would have assumed that if the props.value is the same as what is in the textform (and thus why you call this.props.value here instead of this.state.value), we could obtain the value from above.

src/components/distributions/editor/TextForm/TextInput.js
mmcdermott: No, it doesn't right now.

src/components/distributions/editor/TextForm/TextInput.js
mmcdermott: that state doesn't exist atm.

src/components/distributions/editor/TextForm/TextInput.js
mmcdermott: Yes, it is needed. This is used when the whole form unblurs during dismount (like you click outside the metric card).

PR #451, opened 2016-06-01
Only perform blur when text input has focus

OAGr:


src/components/distributions/editor/TextForm/TextInput.js
```diff
@@ -20,6 +20,13 @@ class TextInputEditor extends Component {
     this._onChange(newEditorState)
   }
 
+  componentWillUnmount() {
+    const selection = this.state.editorState.getSelection()
+    if (selection && selection.getHasFocus()){
+      this.props.onBlur()
```
mmcdermott: Why did you take away passing the value up here? If I'm not mistaken, this will mean that on unmount the guesstimate won't ever save. This doesn't seem right to me.

mmcdermott: Nevermind, I hadn't seen the modification you made in TextForm

PR #455, opened 2016-06-01
Begin passing updated_at back up.

mmcdermott:
Needs https://github.com/getguesstimate/guesstimate-server/pull/98


OAGr:
This turns back on conflict resolution.  Test first

PR #461, opened 2016-06-03
Minor editor enhancements

OAGr:
- Escape on metric value field selects metric.
- Mouse must move to make multiple selection.
- Tab in guesstimate form makes new guesstimate to the right.  Enter makes one downward.

### DIFF
--- a/src/components/lib/FlowGrid/FlowGrid.jsx
+++ b/src/components/lib/FlowGrid/FlowGrid.jsx
@@ -15,6 +15,9 @@ import './FlowGrid.css'
 
 let upto = (n) => Array.apply(null, {length: n}).map(Number.call, Number)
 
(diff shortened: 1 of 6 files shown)


src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -167,6 +178,18 @@ export default class FlowGrid extends Component{
     return Math.max(6, lowestItem, selected) || 6;
   }
 
+  _onReturn(l){
```
mmcdermott: no single letter variable names please. Especially with l, which looks like a 1.

OAGr:
Replaced for now with this: 
https://github.com/getguesstimate/guesstimate-app/pull/462

PR #462, opened 2016-06-03
Minor editor enhancements #2

OAGr:
- Escape on metric value field selects metric.
- Mouse must move to make multiple selection.

### DIFF
--- a/src/components/lib/FlowGrid/FlowGrid.jsx
+++ b/src/components/lib/FlowGrid/FlowGrid.jsx
@@ -15,6 +15,9 @@ import './FlowGrid.css'
 
 let upto = (n) => Array.apply(null, {length: n}).map(Number.call, Number)
 
+//It would be better to have this as state, but we don't want this to cause renders.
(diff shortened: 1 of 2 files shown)


src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -71,14 +74,22 @@ export default class FlowGrid extends Component{
...
+  _mouseMoved(e){
+    const sameLocation = (e.pageX === lastMousePosition.pageX) && (e.pageY === lastMousePosition.pageY)
+    return !sameLocation
+  }
+
+  _handleCellMouseEnter(location, e) {
     if (this.state.leftDown) {
-      this.setState({hover: {row: -1, column: -1}})
-      this._handleEndRangeSelect(location)
+      if (this._mouseMoved(e)) {
```
mmcdermott: Just do &&. 
```
if this.state.leftDown && this._mouseMoved(e) {
...
```

PR #463, opened 2016-06-03
Saving data in the text form should run simulations.

mmcdermott:
Also changes from `*.map(` syntax to `_.map(*, ` syntax to avoid SI infinite recursion errors in dev.


src/components/distributions/editor/index.js
```diff
@@ -29,11 +29,11 @@ export default class Guesstimate extends Component{
   _handleChange(params, runSimulations=true, registerGraphChange=false) {
     this.props.dispatch(changeGuesstimate(this.props.metricId, {...this.props.guesstimate, ...params}, runSimulations, registerGraphChange))
   }
-  _handleSave(params) {
-    if (!_.isEmpty(params)) {this._handleChange(params, false, true)}
+  _handleSave(params, runSimulations=false) {
```
OAGr: why can't runSimulations be in params?

PR #469, opened 2016-06-04
Points should get statistics too, for rendering.

mmcdermott:
We'll short circuit and assign them manually, though, to avoid wasted work in this common edge case.


src/modules/simulations/reducer.js
```diff
@@ -2,8 +2,16 @@ import e from 'lib/engine/engine'
 import {sampleMean, sampleStdev, percentile, cutoff, sortDescending} from 'lib/dataAnalysis.js'
 
 function addStats(simulation){
```
OAGr: I'd feel more comfortable if this had a comment above saying it was mutating.

PR #471, opened 2016-06-04
Only render metric token if necessary. Also mild refactors.

mmcdermott:


mmcdermott:
The CI failure is illusory. Codeship not recognizing test completion.

PR #484, opened 2016-06-05
Refactor the distribution text editor.

mmcdermott:
This clarifies what operations cause data to be saved, and limit information transfer only to what is strictly necessary.


OAGr:
I like this, seems quite sensible.

PR #485, opened 2016-06-06
Add performance timers in the development client.

mmcdermott:
This records how long a variety of components take to render, how long the denormalized space selector takes to run, and when reducers are called.

### DIFF
--- a/src/components/lib/FlowGrid/FlowGrid.jsx
+++ b/src/components/lib/FlowGrid/FlowGrid.jsx
@@ -209,15 +209,26 @@ export default class FlowGrid extends Component{
     )
   }
 
+  componentDidMount() {
+    if (__DEV__) { window.RecordMountEvent(this) }
+  }
+
+  componentWillUpdate() {
+    if (__DEV__) { window.RecordRenderStartEvent(this) }
+  }
+
   componentDidUpdate() {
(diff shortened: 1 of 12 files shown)


src/routes/app.js
```diff
@@ -40,6 +40,89 @@ app.extend({
 
     if (__DEV__) {
       window.Perf = require('react-addons-perf')
+      window.Paused = false
+      window.ClearRecordings = () => {
```
OAGr: I think there should be an object here.  Like, window.guesstimateRecorder or something.  These commands should be on that recorder.  Tying them all directly to widow seems messy to me.

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -209,15 +209,26 @@ export default class FlowGrid extends Component{
     )
   }
 
+  componentDidMount() {
+    if (__DEV__) { window.RecordMountEvent(this) }
```
OAGr: This isn't super Dry.  Maybe in a module is on the top level, then we call that from places like this.

`` import {recorder} ``
``recorder.RecordMountEvent(this)`` (the IF DEV) happens in that module, not everywhere else)

PR #488, opened 2016-06-07
Enter & Tab based navigation.

mmcdermott:


src/components/distributions/editor/index.js
```diff
@@ -65,6 +65,24 @@ export default class Guesstimate extends Component{
...
+      this.props.jumpSection()
+    } else {
+      this.props.onReturn()
+    }
+    return true
+  }
+
+  handleTab(shifted) {
+    if (shifted) {
+      this.props.jumpSection()
```
OAGr: do shift-return and shift-tab do the same thing?

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -176,6 +176,22 @@ export default class FlowGrid extends Component{
     return Math.max(6, lowestItem, selected) || 6;
   }
 
+  _onReturn(l, down){
```
OAGr: make 'l' 'location'

src/components/lib/FlowGrid/FlowGrid.jsx
```diff
@@ -176,6 +176,22 @@ export default class FlowGrid extends Component{
...
+    const {row, column} = l
+    const newRow = down ? row + 1 : (row || 1) -1
+    const newLocation = {row: newRow, column}
+    this.props.onAddItem(newLocation)
```
OAGr: Do we still need to AddItem?  I feel like it's not needed when there's an item there.  Maybe there should be a combined action for these two.

src/components/metrics/card/index.js
```diff
@@ -70,16 +70,27 @@ export default class MetricCard extends Component {
...
     const hasContent = this.refs.MetricCardViewSection.hasContent()
     if (!this.props.inSelectedCell && this._isEmpty() && !hasContent && !this.state.modalIsOpen){
       this.handleRemoveMetric()
     }
+    if (!prevProps.inSelectedCell && this.props.inSelectedCell && !!this.props.selectedFrom) {
+      if (this.props.selectedFrom === 'UP' || this.props.selectedFrom === 'LEFT') {
+        this.refs.MetricCardViewSection.focusName()
+      } else if (this.props.selectedFrom === 'DOWN' || this.props.selectedFrom === 'RIGHT') {
+        this._focusForm()
+      }
```
OAGr: These few lines above should be cleaned up.

src/components/metrics/card/name/index.js
```diff
@@ -107,6 +98,25 @@ export default class MetricName extends Component {
...
+    return true
+  }
+
+  onTab(e) {
+    if (e.shiftKey) {
+      this.props.onTab(false)
+    } else {
+      this.props.jumpSection()
+    }
+    e.preventDefault()
```
OAGr: Same comment from earlier.

src/components/distributions/editor/index.js
mmcdermott: Yes, they do in this case, but it felt clearer to me to have them separated based on whether it was a tab or a return rather than whether it was shifted or not.

OAGr: I played with it, functionality works fine.

PR #492, opened 2016-06-09
Fix up background layer, row heights calculation, and row height during drag to improve performance and UI on editor.

mmcdermott:
This also makes the dragged cell keep the height of the row at the time of the drag, both to help the background rendering issue and to prevent the height from jiterring.

### DIFF
--- a/src/components/lib/FlowGrid/background-container.js
+++ b/src/components/lib/FlowGrid/background-container.js
@@ -7,6 +7,8 @@ import GridPoint from './gridPoints'
 
 import {PTRegion} from 'lib/locationUtils'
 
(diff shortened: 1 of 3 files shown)


src/components/lib/FlowGrid/background-container.js
```diff
@@ -26,7 +26,7 @@ export class BackgroundContainer extends Component {
     copiedRegion: PTRegion,
   }
 
-  shouldComponentUpdate(nextProps) {
+  shouldComponentUpdate(nextProps, nextState) {
```
OAGr: change isn't needed

PR #495, opened 2016-06-09
Adds the ability to import slurps from the hard URL '/import_from_slurp' or from the File menu within a space.

mmcdermott:


OAGr:
Fails several tests

PR #502, opened 2016-06-09
Adds the ability to import slurps via the file menu on the model page

mmcdermott:
This works on the example slurp file, once you correct for the starting '.' cases.

On stage.getguesstimate.com


src/routes/router.js
```diff
@@ -58,6 +59,7 @@ export default Router.extend({
     'scratchpad': 'scratchpad',
     'pricing': 'pricing',
     'subscribe/:id': 'subscribe',
+    'import_from_slurp': 'importFromSlurp',
```
OAGr: This should be removed at this point.

src/lib/slurpParser.js
```diff
@@ -0,0 +1,29 @@
+import e from 'gEngine/engine'
+
+export function parseSlurp(slurp, space={metrics: [], guesstimates: []}) {
```
OAGr: Can this be tested?

src/lib/slurpParser.js
```diff
@@ -0,0 +1,29 @@
...
+  const name = slurp.name
+  const description = slurp.provenance
+  
+  const maxRow = space.metrics.map(m => m.location.row).reduce((x,y) => Math.max(x,y), 0) + 1
+
+  let existingReadableIds = space.metrics.map(m => m.readableId)
+  const width = 8
```
OAGr: where does this 8 come from?

src/components/import_from_slurp/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+@connect()
+export class ImportFromSlurp extends Component {
+  onImport(slurp) {
+    const spaceParams = parseSlurp(slurp)
+    const space = {
+      ..._.pick(spaceParams, ['name', 'description']),
+      graph: {metrics: spaceParams.newMetrics, guesstimates: spaceParams.newGuesstimates}
+    }
+    this.props.dispatch(create(null, space))
```
OAGr: Why does the Slurp form have to do this dispatch?

src/components/import_from_slurp/import_from_slurp_form.js
```diff
@@ -0,0 +1,43 @@
+import React, {Component} from 'react'
```
OAGr: The import_from_slurp component should live within the space, probably the space show.

src/routes/router.js
mmcdermott: Are you sure we don't want a way to import standalone models? Are most slurps complete sets, or designed to be analyzed only with other slurps?

OAGr:
Two changes:
1) The 'Import' field should be greyed out if this isn't valid.
2) There should be a question mark next to the 'Import' button to bring up documentation.

mmcdermott:
Comments addressed; help article is also live.

src/components/spaces/show/Toolbar/import_from_slurp_form.js
```diff
@@ -0,0 +1,62 @@
...
+
+export class ImportFromSlurpForm extends Component {
+  state = {
+    value: ""
+  }
+
+  componentDidMount() {
+    // TODO(Ozzie): Ozzie, any idea how to get this to focus? I can't seem to...
+    this.refs.textarea.focus()
+    ReactDOM.findDOMNode(this.refs.textarea).focus()
```
OAGr: Can this just be removed?  It doesn't matter much, and I don't want it here if it is not working.

src/components/spaces/show/Toolbar/index.js
```diff
@@ -85,8 +97,31 @@ export class SpaceToolbar extends Component {
...
+        left: '30%',
+        width: '40%',
+        right: 'auto',
+        bottom: 'auto',
+        marginRight: '-50%',
+        backgroundColor: '#F0F0F0',
+        border: 'none',
+        padding: '1em'
+      }
+    }
```
OAGr: This is kind of messy, but we can refactor it when n=2

src/components/spaces/show/index.js
```diff
@@ -121,6 +124,19 @@ export default class SpacesShow extends Component {
     this.props.dispatch(spaceActions.destroy(this.props.denormalizedSpace))
   }
 
+  onImportSlurp(slurpObj) {
+    const space = this.props.denormalizedSpace
+
+    const spaceUpdates = parseSlurp(slurpObj, space)
+    let noGraphUpdates = {}
```
OAGr: what does noGraphUpdates mean?  Change to another var name

src/components/spaces/show/index.js
```diff
@@ -121,6 +124,19 @@ export default class SpacesShow extends Component {
...
+    if (!space.name) {noGraphUpdates.name = spaceUpdates.name}
+    if (!space.description) {noGraphUpdates.description = spaceUpdates.description}
+    this.props.dispatch(spaceActions.update(this._id(), noGraphUpdates))
+    this.props.dispatch({type: 'ADD_METRICS', items: spaceUpdates.newMetrics, newGuesstimates: spaceUpdates.newGuesstimates})
+    this.props.dispatch(spaceActions.updateGraph(this._id()))
+    this.props.dispatch(simulationActions.runSimulations(this._id(), spaceUpdates.newMetrics))
```
OAGr: This to me feels like it should all be handled from one action, which is the ImportSlurp action.  I realize that doing so would result in one extra render, but it feels more succinct to me.

src/components/spaces/show/index.js
```diff
@@ -121,6 +124,19 @@ export default class SpacesShow extends Component {
...
   }
 
+  onImportSlurp(slurpObj) {
+    const space = this.props.denormalizedSpace
+
+    const spaceUpdates = parseSlurp(slurpObj, space)
+    let noGraphUpdates = {}
+    if (!space.name) {noGraphUpdates.name = spaceUpdates.name}
+    if (!space.description) {noGraphUpdates.description = spaceUpdates.description}
+    this.props.dispatch(spaceActions.update(this._id(), noGraphUpdates))
```
OAGr: I imagine that this dispatch should only be done if noGraphUpdates is not empty.

src/lib/slurpParser-test.js
```diff
@@ -0,0 +1,63 @@
...
+    if (!matchPresent(pattern[key],sample[key])) {
+      console.warn("Objects differ at key ", key)
+    }
+  }
+  return matches
+}
+
+describe('parseSlurp', () => {
+  describe('Valid Slurp', () => {
+    const spaceName = "Space"
```
OAGr: single quotes for this stuff

src/lib/slurpParser.js
```diff
@@ -0,0 +1,30 @@
+import e from 'gEngine/engine'
+
+const MAX_WIDTH = 8
```
OAGr: Maybe rename this MAX_COLUMN, with a comment to say that this is a usability feature so that lots of SIPs in a SLURP wouldn't be frustratingly long?

src/components/spaces/show/index.js
mmcdermott: Yeah, I would say this plus the formula fill actions makes me really want to do a widespread refactor, which I'd rather do when those both are in, so I'd want to leave this for now.

PR #506, opened 2016-06-09
Fixes some jitteriness with background.

mmcdermott:
If the state is calculated in Flowgrid, it is calculated on the previous incarnation of the rendered component. If it is calculated in the background container, that happens after the rows are rendered are thus gets the correct row heights.

### DIFF
--- a/src/components/lib/FlowGrid/background-container.js
+++ b/src/components/lib/FlowGrid/background-container.js
@@ -7,6 +7,8 @@ import GridPoint from './gridPoints'
 
 import {PTRegion} from 'lib/locationUtils'
 
+const upto = (n) => Array.apply(null, {length: n})
+
 const Region = ({rowHeights, columnWidth, selectedRegion, type}) => {
   const gridPoint = new GridPoint({rowHeights, columnWidth, padding: 0})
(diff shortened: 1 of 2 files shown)


src/components/lib/FlowGrid/background-container.js
```diff
@@ -20,23 +22,36 @@ export class BackgroundContainer extends Component {
...
+    this.setState({rowHeights: _.map(upto(nextProps.rowCount), (r, i) => nextProps.getRowHeight(i))})
+  }
+
   shouldComponentUpdate(nextProps, nextState) {
     return (
       !_.isEqual(this.props.copiedRegion, nextProps.copiedRegion) ||
       !_.isEqual(this.props.selectedRegion, nextProps.selectedRegion) ||
       !_.isEqual(this.props.edges, nextProps.edges) ||
-      !_.isEqual(this.props.rowHeights, nextProps.rowHeights)
+      !_.isEqual(this.state.rowHeights, nextState.rowHeights)
```
OAGr: this seems sketchy to me.  I'm not sure if _.isEqual works for arrays, though we are using it for arrays and objects above.

PR #508, opened 2016-06-10
Organization creation improvements

OAGr:
This does a few things:

- Organizations have default images
- Organization spaces on the free plans cannot be made private.
- The free trail information is styled green.
- Once a member is removed from an organization, the corresponding invitation will be deleted as well.

Merge this first:
https://github.com/getguesstimate/guesstimate-server/pull/106


src/modules/organizations/actions.js
```diff
@@ -41,15 +41,15 @@ export function fetchById(organizationId) {
...
 export function fetchSuccess(organizations) {
   return (dispatch) => {
-    const formatted = organizations.map(o => _.pick(o, ['id', 'name', 'picture', 'admin_id', 'account']))
+    const formatted = organizations.map(o => _.pick(o, ['id', 'name', 'picture', 'admin_id', 'account', 'plan']))
     dispatch(oActions.fetchSuccess(formatted))
   }
 }
 
-export function create(name) {
+export function create({name, planId}) {
```
mmcdermott: Kind of a weird API here; taking an object with signature {name, planId} just to use an object later with signature {name, plan}. Why not just take the organization object directly in the create.

PR #532, opened 2016-06-16
Center metric modal by increasing width slightly. Closes 226.

mmcdermott:


mmcdermott:
This will be terrible on large monitors. Closing and reconsidering.

PR #534, opened 2016-06-17
Tracks Elev usage via segment. Closes 286.

mmcdermott:
Uses a set timeout b/c creating a jquery handle immediately doesn't detect the element if it hasn't finished loading.


src/server/segment/constants.js
```diff
@@ -11,6 +11,7 @@ export const ACCOUNT_MODAL_CLICK = "Clicked account modal"
 export const USER_MENU_OPEN = "Opened user menu"
 export const USER_MENU_CLOSE = "Closed user menu"
 export const COPIED_MODEL = "Copied model"
+export const USED_ELEV_DOCUMENTATION = "Used elev"
```
OAGr: Maybe 'Opened elev'?

PR #542, opened 2016-06-22
Basic Calculator View

mmcdermott:
Sets up the Calculator show page and the necessary associated backend. No creation just yet.
calculator-view presubmit
calc view submitted


src/components/calculators/calculator-space-selector.js
```diff
@@ -0,0 +1,52 @@
...
+const OUTPUT = 'OUTPUT'
+const INPUT = 'INPUT'
+const NOEDGE = 'NOEDGE'
+
+const relationshipType = (edges) => {
+  if (edges.inputs.length && edges.outputs.length) { return INTERMEDIATE }
+  if (edges.inputs.length) { return OUTPUT }
+  if (edges.outputs.length) { return INPUT }
+  return NOEDGE
+}
```
OAGr: Much of this code is in another selector, it would be nice to clean up.

src/lib/engine/calculator.js
```diff
@@ -0,0 +1,3 @@
+export function fullUrl({id}) {
+  return `https://www.getguesstimate.com/calculators/${id}`
```
OAGr: I feel like the absolute base url should be somewhere else, this seems quite messy to me.

mmcdermott:
Addressed comments, and re-staged to stage.getguesstimate.com. Note that with the BASE_URL change, facebook share crashes locally (but works on stage).

PR #544, opened 2016-06-23
Fix save error state (and test so this issue does not recur).

mmcdermott:


src/modules/spaces/actions-test.js
```diff
@@ -0,0 +1,25 @@
...
+      .patch('/spaces/1')
+      .reply(422, {})
+
+    const expectedActions = [
+      { type: 'SPACES_UPDATE_START', record: {id: 1}, data: undefined },
+      { type: 'CHANGE_CANVAS_STATE', values: {actionState: 'SAVING'} },
+      { type: 'CHANGE_CANVAS_STATE', values: {actionState: 'ERROR'} },
+    ]
+    const store = mockStore({ spaces: [{id: 1}] }, expectedActions, done)
+    store.dispatch(actions.generalUpdate(1))
```
OAGr: store.dispatch doesn't sound like a test to me, it reads like set up.

PR #545, opened 2016-06-23
You should be logged out visibly when your token expires.

mmcdermott:
A bit hacky, but the auth0 documentation was unyielding in providing a better solution.


mmcdermott:
Fixed up a bit; should be more comprehensive now

OAGr:
This section of the code base is getting confusing, but let's do this for now.

PR #551, opened 2016-06-26
The calculator should always show the current and correct answer.

mmcdermott:
This clears the old simulations when the calculator loads, and ensures that the calculator only re-simulates when all inputs are filled, and only shows the 'calculate' button when the results are present, to avoid showing an empty output row.

Adds a small amount of lag to the calculate button, and some cruft in the simulation pathway I want to clean up later.

On stage.getguesstimate.com

### DIFF
--- a/src/components/calculators/show.js
+++ b/src/components/calculators/show.js
@@ -64,16 +64,13 @@ export class CalculatorShow extends Component {
   }
 
   onEnter() {
-    if (!this.state.showResult && this.readyToCalculate()) {
-      this.setState({showResult: true})
-    }
-    return true
+    if (!this.state.showResult && this.readyToCalculate()) { this.setState({showResult: true}) }
   }
 
   allInputsHaveContent(idsToExclude=[]) {
     const includedInputs = this.props.inputs.filter(i => !_.some(idsToExclude, id => i.id === id))
     const inputComponents = _.map(includedInputs, metric => this.refs[`input-${metric.id}`])
-    return inputComponents.map(i => !!i && i.hasValidContent()).reduce((x,y) => x && y, true)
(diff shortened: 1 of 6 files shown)


src/components/calculators/show.js
```diff
@@ -48,20 +50,40 @@ export class CalculatorShow extends Component {
...
+    if (shouldRunSims) {this.setState({hasSimulated: true})}
   }
 
-  allInputsHaveContent() {
-    const inputComponents = _.map(this.props.inputs, metric => this.refs[`input-${metric.id}`])
+  onEnter() {
+    if (!this.state.showResult && this.readyToCalculate()) {
+      this.setState({showResult: true})
+    }
+    return true
```
OAGr: I feel like this shouldn't have to know about the 'return true' statement.  That one element should be reserved to the input.

src/components/calculators/show.js
```diff
@@ -48,20 +50,40 @@ export class CalculatorShow extends Component {
...
+  }
+
+  allInputsHaveContent(idsToExclude=[]) {
+    const includedInputs = this.props.inputs.filter(i => !_.some(idsToExclude, id => i.id === id))
+    const inputComponents = _.map(includedInputs, metric => this.refs[`input-${metric.id}`])
     return inputComponents.map(i => !!i && i.hasValidContent()).reduce((x,y) => x && y, true)
   }
 
+  allOutputsHaveStats() {
+    return this.props.outputs.map(o => !!o && _.has(o, 'simulation.stats')).reduce((x,y) => x && y, true)
```
OAGr: does this ``reduce`` statement do the same as a _.some() statement?

PR #556, opened 2016-06-29
Misc. Refactors centered around refining the change-guesstimate and simulation actions

mmcdermott:


src/lib/recorder.js
```diff
@@ -16,70 +16,88 @@ export class GuesstimateRecorder {
 
   constructor() {
     this.disabled = !__DEV__
-    this.paused = true
+    this.paused = false // TODO(matthew): change back.
```
OAGr: ?  TODO Before PR?

src/modules/simulations/actions.js
```diff
@@ -7,17 +7,10 @@ import {GraphPropagation} from '../../lib/propagation/graph-propagation'
...
-    const metric = e.metric.get(getState().metrics, metricId)
-    const spaceId = !!metric && metric.space
-    const propagation = new GraphPropagation(dispatch, getState, {spaceId, onlyUnsimulated: true})
-    yield propagation.run()
-  } else {
-    const propagation = new GraphPropagation(dispatch, getState, {metricId, onlyHead: true})
-    yield propagation.run()
-    yield* runFormSimulation({getState, metricId, dispatch})
-  }
+export function* runFormSimulation({getState, metricId, dispatch}) {
```
OAGr: runFormSimulation only runs the head?  I wouldn't have guessed that.

src/modules/simulations/actions.js
```diff
@@ -7,17 +7,10 @@ import {GraphPropagation} from '../../lib/propagation/graph-propagation'
...
+  const propagation = new GraphPropagation(dispatch, getState, {metricId, onlyHead: true})
```
OAGr: why do onlyHead then run descendants?

mmcdermott: We do this in canvas editing so that resimulating the card being edited is quick, but the rest of the chain still updates later. I'd rather not pull out those two lines into a separate function at this point, as I don't ever anticipate a need for that functionality to be used separately.

PR #559, opened 2016-07-04
Initial fact bank setup.

mmcdermott:
The classes `noun`, `property`, and `suggestion` need further styling, I'm sure, and there will likely be more refactors, but this is a first pass. And we'll want to re-format the list of cities. But there's a lot here, so I thought it worth throwing your way for comments in the meantime.


OAGr:
While cleaning the cities, could everything be removed except for name and population, for now (unless we could get other stats easily)

mmcdermott:
I left all the numeric properties. The growth rates at the very least may be useful. If you disagree, I'm happy to remove them, but it seems wise to have at least two properties per city, as that will be a more representative example of the eventual product. I'm closing this PR to open one off the other branch.

PR #585, opened 2016-07-11
Fix the autofill bug.

mmcdermott:


src/modules/auto_fill_region/actions-test.js
```diff
@@ -34,5 +34,20 @@ describe('Autofill Actions', () => {
...
+        {
+          id: '2',
+          readableId: 'VL',
+          location: {row: 0, column: 0},
+        },
+      ]
+      const fillFn = actions.fillDynamic(startMetric, startGuesstimate, direction)
+      const {metric, guesstimate} = fillFn(location, metrics)
+      console.log(metric, guesstimate)
+      expect(!metric && !guesstimate).to.equal(true)
```
OAGr: Do you need the console.log?  
Also, the next line is confusing. Maybe something more like,
``expect(!!guesstimate).to.equal(false)``
``expect(!!metric).to.equal(false)``

Although to be honest I'm really not sure what is happening there.

src/modules/auto_fill_region/actions.js
```diff
@@ -44,7 +44,10 @@ export function fillDynamic(startMetric, startGuesstimate, direction) {
       const matchedMetric = metrics.find(m2 => isAtLocation(translateFn(m.location), m2.location))
       if (!!matchedMetric) { idMap[m.readableId] = matchedMetric.readableId }
     })
-    if (_.isEmpty(idMap)) { return {metric, guesstimate: {...startGuesstimate, metric: metric.id}} }
+    if (_.isEmpty(idMap)) {
+      if (numNonConstantInputs === 0) { return {metric, guesstimate: {...startGuesstimate, metric: metric.id}} }
+      return {}
```
OAGr: I'd probably have an ``else`` statement here.

PR #609, opened 2016-07-14
Adds visible error explanations for certain classes of errors. Closes 590.

mmcdermott:
Staged on matthew.getguesstimate.com

Remaining AIs:
1. Ensure error messages are acceptable.
2. z-Index for error tooltip (and reposition to be below the card)
3. Prevent INTERNAL_ERRORS from printing error messages (though this may not be necessary, as these rarely, if ever, occur).


OAGr:
- Error tooltip needs to be brought up to the metric card file in order to resolve z-index issue.
- Error tooltip is broken in metric modal.  Fixing the above issue should help.  I think for now we shouldn't show the error in the modal.
- Error should show up whenever field is selected, not just on hover.  It should also show as the user types.  Right now it seems to only show after the user is done typing and re-selects that area.

PR #615, opened 2016-07-15
Adds decorators that wrap all valid and erroneous inputs with different styles. Closes 588 & 589

mmcdermott:
node-name-highlight

Also on stage-matthew.getguesstimate.com


mmcdermott:
Test failure is illusory, all tests actually pass.

OAGr:
One bug: This crashes when one clicks the 'expand' button on a metric.

PR #617, opened 2016-07-15
Added an 'expanded view' to show all inputs expanded and all readable IDs. Closes 573.

mmcdermott:
I'm leaving the image for this view mode up to you to add, Ozzie, as I presumed you would want to take that. If not, and you want me to take a stab at it, happy to, just let me know.
expanded-view


OAGr:
Good work.  One important point though; this should not be a view option, but rather, an independent toggle.  We're going to translate the view options into independent toggles, as they are quite independent.  

It should work as a button next to the undo/redo arrows, that can be toggled.  I can handle styling.

OAGr:
Closing PR for now, reopen when you fix that.

PR #618, opened 2016-07-16
Adds a mode to show all inputs and makes the view options toggles. Closes 573.

mmcdermott:
Also fixes a bug about `tbody` in scientific view.


OAGr:
This seems weird to me, where is the background? 

image

Seems like it's in 'display mode'

mmcdermott:
Corrected and re-staged at stage-matthew.getguesstimate.com

mmcdermott:
Known issue: Height sticks after engaging scientific view.

OAGr:
- [ ] Grid lines in background don't show.
- [ ] Highlight around selected arrows option doesn't show.
- [ ] /embed shows grid lines, but should not show them.
- [ ] You may want to add another option for 'gridlines' to hide or show those.

OAGr:
This is on hold for me to make the correct icon.

PR #656, opened 2016-07-21
Calculator fixes

OAGr:
- Made input descriptions use Markdown.
- Added toggle for a stats description for the outputs.

toggle

### DIFF
--- a/src/components/calculators/show/output.js
+++ b/src/components/calculators/show/output.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {Component} from 'react'
 
 import numberShow from 'lib/numberShower/numberShower'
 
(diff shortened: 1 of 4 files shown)


src/components/calculators/show/output.js
```diff
@@ -9,27 +10,69 @@ const PrecisionNumber = ({value, precision, number=numberShow(value, precision)}
...
 
-const RangeDisplay = ({range: [low, high]}) => (
+const RangeDisplay = ({low, high}) => (
   <div><PrecisionNumber value={low}/> to <PrecisionNumber value={high}/></div>
 )
 
-const ResultSection = ({length, mean, adjustedConfidenceInterval}) => (
-  length === 1 ? <PrecisionNumber value={mean} precision={6}/> : <RangeDisplay range={adjustedConfidenceInterval}/>
+const ResultSection = ({length, mean, percentiles}) => (
+  length === 1 ? <PrecisionNumber value={mean} precision={6}/> : <RangeDisplay low={percentiles[5]} high={percentiles[95]}/>
```
mmcdermott: Why this change?

src/components/calculators/show/output.js
```diff
@@ -9,27 +10,69 @@ const PrecisionNumber = ({value, precision, number=numberShow(value, precision)}
...
-    <div className='row'>
-      <div className='col-xs-12 col-sm-7'>
-        <div className='name'>
-          {name}
-        </div>
-      </div>
-      <div className='col-xs-12 col-sm-5'>
-        <div className={`result-section${!_.isEmpty(_.get(simulation, 'sample.errors')) ? ' has-errors' : ''}`}>
-          {_.has(simulation, 'stats') && <ResultSection {...simulation.stats} />}
+const AnalyticsSection = (stats) => {
```
mmcdermott:
```
const AnalyticsSection = (stats) => (
  <div className='stats-summary'>
  ...
)
```
(e.g. no need for `{ return (` with es6 form functions)

src/components/calculators/show/output.js
```diff
@@ -9,27 +10,69 @@ const PrecisionNumber = ({value, precision, number=numberShow(value, precision)}
...
+    </div>
+  )
+}
+
+export class Output extends Component {
+  state = {
+    showAnalysis: false
+  }
+
+  showAnalysis(show=true) {
```
mmcdermott: I would either call this `toggleAnalysis()` and have it just flip the bit, or have two one-liner functions for `showAnalysis()` and `hideAnalysis()`, but it's not that important. I only mention it b/c I had other comments anyways.

src/components/calculators/show/output.js
```diff
@@ -9,27 +10,69 @@ const PrecisionNumber = ({value, precision, number=numberShow(value, precision)}
...
+          <div className='col-xs-12 col-sm-7'>
+            <div className='name'>
+              {name}
+            </div>
+          </div>
+          <div className='col-xs-12 col-sm-5'>
+            <div className={`result-section${!_.isEmpty(_.get(simulation, 'sample.errors')) ? ' has-errors' : ''}`}>
+              {_.has(simulation, 'stats') && <ResultSection {...simulation.stats} />}
+
+              {!showAnalysis && _.has(simulation, 'stats.percentiles.5') &&
```
mmcdermott: Let's talk about this a bit; I don't think you should be using the percentiles here, I think you should be using the 95% CI.

mmcdermott: I would probably find this a bit cleaner if you declared a var in the render with the icon content, then used it here.

PR #665, opened 2016-07-28
Replaced favicon, removed old logo

OAGr:


mmcdermott:
This looks like it deletes the logo from the header, too? Why? Also, why keep the old favicon still around as a file? It will be present in version control.

OAGr:
It deletes it because I didn't think this newer logo actually looked that great next to our name, and I thought it looked fine without it.

I don't mind keeping that file around, just in case we do want to modify it soon later.  It's not very big.

image

PR #671, opened 2016-08-01
First pass at the organization fact tab.

mmcdermott:


src/modules/organizations/actions.js
```diff
@@ -75,16 +80,24 @@ export function addMember(organizationId, email) {
...
         dispatch(membershipActions.createSuccess([membership]))
       }
     })
   }
 }
+
+export function addFact(organization, fact) {
+  return (dispatch, getState) => {
+    const cid = cuid()
+    let object = {id: cid, organization_id: organization.id, user_id: 4}
```
OAGr: why is the user_id 4? Is that an accident?

src/routes/router.js
```diff
@@ -85,7 +88,7 @@ export default Router.extend({
   faq() { this.render(<FAQ/>) },
   subscribe(id) { this.render(<FirstSubscriptionPage planName={id}/>) },
   userShow(id) { this.render(<UserShow userId={id}/>, {backgroundColor: 'GREY'}) },
-  organizationShow(id) { this.render(<OrganizationShow organizationId={id} key={id}/>, {backgroundColor: 'GREY'}) },
+    organizationShow(id) { this.render(<OrganizationShow organizationId={id} key={id} tab={extractOrganizationTab(window.location.search)}/>, {backgroundColor: 'GREY'}) },
```
OAGr: This is a pretty hacky way of doing this.  I believe there's a better way with redux router.  That said, it isn't that great either.  

In the least, I'd' hide 'window.location.search') within the function ``extractOrganizationTab``, which ideally would call a third function, ``extractSearchParams`` or similar.

src/routes/router.js
```diff
@@ -30,6 +30,9 @@ import configureStore from './middleware'
 
 import * as segment from 'servers/segment/index'
 
+const ORGANIZATION_TAB_REGEX = /tab=(\w+)/
```
OAGr: I believe it's more correct to use ?tab=, rather than /tab, for what is effectively a parameter instead of a query (indicated by the = sign)

src/components/organizations/show/facts.js
```diff
@@ -0,0 +1,161 @@
...
+        </div>
+      </div>
+      <div className='col-md-1'>
+        <span className='ui button options'><Icon name='ellipsis-v' /></span>
+      </div>
+    </div>
+  </div>
+)
+
+const hasAllNonEmpty = (obj, props) => !_.some(props.map(prop => _.isEmpty(_.get(obj, prop))))
```
OAGr: why not ``_.all`` instead? (but the inverse)

this could also be split into two lines or something, quite terse.

src/components/organizations/show/facts.js
```diff
@@ -0,0 +1,161 @@
...
+const readableIdPartFromWord = word => (/\d/).test(word) ? word : word[0]
+function getVariableNameFromName(rawName) {
+  const name = rawName.trim().replace(/[^\w\d]/g, ' ').toLowerCase()
+  const words = name.split(/[^\w\d]/).filter(s => !_.isEmpty(s))
```
OAGr: a lot of this logic seems to be similar to other stuff I've seen before. Is it?  Can this be dried up at all?

src/components/organizations/show/facts.js
```diff
@@ -0,0 +1,161 @@
...
+  if (words.length === 1 && name.length < 10) {
+    return name
+  } else if (words.length < 3) {
+    return name.slice(0,3)
+  } else {
+    return words.map(readableIdPartFromWord).join('')
+  }
+}
```
OAGr: does this handle conflicts of words?  I'm assuming not.  It should.

for instance, this could be resolved as something basic as 'FoB1', 'FoB2' and so on.

src/routes/router.js
mmcdermott: that's just part of the regex. It is just checking the `tab=(\w+)` plus bit.. Also, this is just included for development, so hot reloading doesn't change tabs. Sorry for that not being clear, that was present in the original PR description. I was planning on removing this in prod as it is hacky.

src/routes/router.js
mmcdermott: This is just included for development, so hot reloading doesn't change tabs. Sorry for that not being clear, that was present in the original PR description. I was planning on removing this in prod as it is hacky.

src/components/organizations/show/facts.js
```diff
@@ -0,0 +1,161 @@
...
+    this.setFactState(this.state.variableNameManuallySet ? {name} : {name, variable_name: getVariableNameFromName(name)})
+  }
+  onChangeVariableName(e) { this.setFactState({variable_name: _.get(e, 'target.value')}, {variableNameManuallySet: true}) }
+  onChangeExpression(e) { this.setFactState({expression: _.get(e, 'target.value')}) }
+
+  onBlurExpression() {
+    simulateFact(['biz'], this.state.fact).then(({values, errors}) => {
+      let simulation = {sample: {values, errors}}
+      addStats(simulation)
+      simulation.sample.sortedValues = null // We don't want to send this to the server for latency reasons.
```
OAGr: I realize we don't send this to the server, but can that be decided in the code that sends it to the server, rather than here?

src/components/organizations/show/facts.js
mmcdermott: Where have you seen this before? I don't recall anything similar being used.

src/lib/engine/facts.js
```diff
@@ -69,3 +79,13 @@ export function addFactsToSpaceGraph({metrics, guesstimates, simulations}, {glob
     simulations: [...simulations, ...factSimulations],
   }
 }
+
+export function simulateFact(selector, fact) {
+  const guesstimate = toGuesstimate(selector, fact)
+  const graph = {metrics: [{id: guesstimate.metric}], guesstimates: [guesstimate]}
+  const metricPropagation = new MetricPropagation(guesstimate.metric, [], 0)
+
+  return metricPropagation.simulate(5000, graph).then(
```
OAGr: I'd make the ``5000`` an external constant, at least on the top of this file.

src/lib/engine/simulation.js
```diff
@@ -1,5 +1,44 @@
 import * as sample from './sample'
 
+import {sampleMean, sampleStdev, percentile, cutoff, sortDescending} from 'lib/dataAnalysis.js'
```
OAGr: I like that this stuff was moved here.

src/lib/engine/organization.js
```diff
@@ -36,3 +36,5 @@ export function organizationMemberships(organizationId, memberships) {
 export function organizationInvitations(organizationId, invitations) {
   return _.filter(invitations, e => sameId(e.organization_id, organizationId))
 }
+
+export const getFactVar = ({id}) => `organization_${id}`
```
OAGr: I'm not a big fan of this name.  Maybe, ``organization_readable_id`` or similar?

src/components/organizations/show/facts.js
OAGr: Just a whole lot of names and splitting for similar things.  That said, you know the specifics more than I do, just checking.

src/components/organizations/show/facts.js
mmcdermott: I agree with you, but I'd like to handle this in a separate PR to the epic branch. This PR is blocking a lot of things, and getting this all set up with tests and such could take a little while. How does that sound?

src/routes/router.js
OAGr: Can that be removed as part of this story?  We'll have to do it some time, seems cleaner to ensure that each story ends up clean.

src/components/organizations/show/facts.js
OAGr: That's fine.  In that case, this should be a separate story.

Blocking isn't as bad when there's only you doing the coding for this stuff at least.

src/routes/router.js
OAGr: Makes sense

src/routes/router.js
mmcdermott: Discussed offline; leaving for now.

src/modules/organizations/actions.js
```diff
@@ -73,18 +78,29 @@ export function create({name, plan}) {
...
-      else if (membership) {
+      if (!!membership) {
         dispatch(userActions.fetchSuccess([membership._embedded.user]))
         dispatch(membershipActions.createSuccess([membership]))
       }
     })
   }
 }
+
+export function addFact(organization, rawFact) {
```
OAGr: what is a rawFact?  Maybe some documentation?

src/modules/organizations/actions.js
```diff
@@ -73,18 +78,29 @@ export function create({name, plan}) {
...
+  return (dispatch, getState) => {
+    const cid = cuid()
+    let object = {id: cid, organization_id: organization.id}
```
OAGr: is object referenced again?  Does this line do anything?

src/modules/organizations/actions.js
```diff
@@ -73,18 +78,29 @@ export function create({name, plan}) {
...
+
+    let fact = Object.assign({}, rawFact)
+    _.set(fact, 'simulation.sample.sortedValues', null)
+
+    api(getState()).organizations.addFact(organization, fact, (err, fact) => {
```
OAGr: I don't like how ``fact`` is already defined above.  Can you name that var a bit different?

mmcdermott: It sets the ids and timestamps on the server, which we want on the client.

PR #675, opened 2016-08-02
Makes guesstimates be stored internally with an `expression` field, which generates a human readable input.

mmcdermott:
Metrics are referenced by `id` within the `expression` field, not `readableId`. The translation to & from human readable `input` form takes place around the user interactions with the Guesstimate and prior to simulation. 

Must go in (and be deployed) after https://github.com/getguesstimate/guesstimate-server/pull/147

### DIFF
--- a/src/lib/engine/guesstimate.js
+++ b/src/lib/engine/guesstimate.js
@@ -90,3 +90,7 @@ export function inputsFromExpressions(metrics) {
   const translateFn = ({expression}) => expression.replace(RegExp(reParts.join('|'), 'g'), match => idMap[match])
   return g => !_.isEmpty(g.input) || _.isEmpty(g.expression) ? g : {...g, input: translateFn(g)}
 }
+
+export function expressionFromInput(input, idMap) {
+  return input.replace(RegExp(Object.keys(idMap).join('|'), 'g'), match => `\$\{${idMap[match]}\}`)
+}
--- a/src/lib/engine/guesstimate.js
+++ b/src/lib/engine/guesstimate.js
@@ -83,14 +83,22 @@ function _inputMetricsWithValues(guesstimate: Guesstimate, dGraph: DGraph): Obje
   return [inputs, _.uniq(errors)]
 }
 
-export function inputsFromExpressions(metrics) {
+// In the `expression` syntax, input metrics are expressed as `${[metric id]}`. To match that in a regex, and translate
+// to it, we need functions that wrap passed IDs in the right syntax, appropriately escaped.
+const expressionSyntaxPad = id => `\$\{${id}\}`
+const escapedExpressionSyntaxPad = id => `\\\$\\\{${id}\\\}`
+
+// Returns a function which takes a guesstimate and returns that guesstimate with an input based on its
+// expression.
+export function expressionToInputFn(metrics) {
   let idMap = {}, reParts = []
-  metrics.forEach( ({id, readableId}) => {reParts.push(`\\\$\\\{${id}\\\}`); idMap[`\$\{${id}\}`] = readableId} )
+  metrics.forEach( ({id, readableId}) => {reParts.push(escapedExpressionSyntaxPad(id)); idMap[expressionSyntaxPad(id)] = readableId} )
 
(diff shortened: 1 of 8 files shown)


src/lib/engine/guesstimate.js
```diff
@@ -82,3 +82,15 @@ function _inputMetricsWithValues(guesstimate: Guesstimate, dGraph: DGraph): Obje
   })
   return [inputs, _.uniq(errors)]
 }
+
+export function inputsFromExpressions(metrics) {
```
OAGr: If this takes ``metrics``, I feel like it should be in ``metric.js``

OAGr: I'd probably rename this inputsFromExpressionsFn or inputsFromExpressionsFunction or something to indicate it returns a function, really didn't expect that.  Maybe a comment on the top.

src/lib/engine/space.js
```diff
@@ -12,10 +12,11 @@ export function get(collection, id){
   return collection.find(i => (i.id === id))
 }
 
-export function subset(graph, spaceId){
+export function subset(graph, spaceId, withInputs = false){
   if (spaceId){
     const metrics = graph.metrics.filter(m => m.space === spaceId)
-    const guesstimates = _.flatten(metrics.map(m => _metric.guesstimates(m, graph)))
+    const rawGuesstimates = _.flatten(metrics.map(m => _metric.guesstimates(m, graph)))
+    const guesstimates = withInputs ? rawGuesstimates.map(_guesstimate.inputsFromExpressions(metrics)) : rawGuesstimates
```
OAGr: I believe this runs _guesstimate.inputsFromExpressions(metrics) n times.  Can you save it above as a ``const``?  Speed aside, that would make it a tad bit more understandable.

src/lib/engine/guesstimate.js
```diff
@@ -82,3 +82,15 @@ function _inputMetricsWithValues(guesstimate: Guesstimate, dGraph: DGraph): Obje
...
+  let idMap = {}, reParts = []
+  metrics.forEach( ({id, readableId}) => {reParts.push(`\\\$\\\{${id}\\\}`); idMap[`\$\{${id}\}`] = readableId} )
+
+  const translateFn = ({expression}) => expression.replace(RegExp(reParts.join('|'), 'g'), match => idMap[match])
+  return g => !_.isEmpty(g.input) || _.isEmpty(g.expression) ? g : {...g, input: translateFn(g)}
```
OAGr: Add parans to make order of execution more obvious.

src/lib/engine/guesstimate.js
```diff
@@ -82,3 +82,15 @@ function _inputMetricsWithValues(guesstimate: Guesstimate, dGraph: DGraph): Obje
   })
   return [inputs, _.uniq(errors)]
 }
+
+export function inputsFromExpressions(metrics) {
+  let idMap = {}, reParts = []
+  metrics.forEach( ({id, readableId}) => {reParts.push(`\\\$\\\{${id}\\\}`); idMap[`\$\{${id}\}`] = readableId} )
```
OAGr: Maybe make functions that return these regex parts and name them on the top of this file?  The idMap is used more than once.  Comments with these would help.

src/lib/engine/space.js
```diff
@@ -47,13 +48,15 @@ export function toDSpace(spaceId, graph) {
...
   dSpace.edges = _dGraph.dependencyMap(dSpace)
 
+  const withInputFn = _guesstimate.inputsFromExpressions(dSpace.metrics)
   dSpace.metrics = dSpace.metrics.map(s => {
     let edges = {}
     edges.inputs = dSpace.edges.filter(i => i.output === s.id).map(e => e.input)
     edges.outputs = dSpace.edges.filter(i => i.input === s.id).map(e => e.output)
     edges.inputMetrics = edges.inputs.map(i => dSpace.metrics.find(m => m.id === i))
-    return Object.assign({}, s, {edges})
+    return {...s, guesstimate: withInputFn(s.guesstimate), edges}
```
OAGr: for readability, I think edges should be second in this,  Also, you may want to split it for multiple lines.

```
return {
  ...s, 
  edges,
  guesstimate: withInputFn(s.guesstimate)
}
```

PR #678, opened 2016-08-02
The new fact form should reset properly after fact creation.

mmcdermott:
Ostensibly this should chain to the actual API call via a promise or something, but this is sufficient for now. Alternatively, if the storage as state bothers you it can be made a class variable that's not part of state instead, that would also work.


src/components/organizations/show/facts/facts.js
```diff
@@ -8,6 +8,7 @@ import {getVar} from 'gEngine/facts'
 export class FactBookTab extends Component {
   state = {
     editingFactId: null,
+    factsAdded: 0,
```
OAGr: I think this should be called ``newFactKey``, which is really what it's for.

PR #682, opened 2016-08-03
Fact bank restyling 

OAGr:
image


mmcdermott:
It seems like you've moved `/o/s/facts` -> `/o/s/facts/list` uniformly here, which would just add another directory to traverse without changing the grouping of files at all. Is that necessary?

OAGr:
I moved it away from /organizations.  The move was in part because the next and corresponding PR is for the space sidebar, for which it makes sense to separate facts from organizations a bit.

src/components/facts/list/container.js
```diff
@@ -0,0 +1,32 @@
...
+
+function mapStateToProps(state) {
+  return {
+    organizations: state.organizations,
+    organizationFacts: state.facts.organizationFacts,
+  }
+}
+
+@connect(mapStateToProps)
+export class FactListContainer extends Component{
```
mmcdermott: This shouldn't have to re-select the organization like this. Instead of passing in the organization Id, pass the whole org and you can skip that selection.

src/components/facts/list/container.js
```diff
@@ -0,0 +1,32 @@
...
+  displayName: 'FactListContainer'
+
+  render() {
+    const {organizationId, organizations, organizationFacts} = this.props
+    const facts = _.get(organizationFacts.find(f => f.variable_name === `organization_${organizationId}`), 'children') || []
```
mmcdermott: It's probably worth pulling this into a selector.

src/components/facts/list/container.js
```diff
@@ -0,0 +1,32 @@
...
+    const organization = organizations.find(u => u.id.toString() === organizationId.toString())
+    return (
+      <FactList
+        onDeleteFact={fact => this.props.dispatch(organizationActions.deleteFact(organization, fact))}
+        onAddFact={fact => this.props.dispatch(organizationActions.addFact(organization, fact))}
```
mmcdermott: Does this have the problem of the new form maintaining old entries still? It looks like you dropped my change that fixed that.

src/components/facts/list/form.js
```diff
@@ -75,51 +73,54 @@ export class FactForm extends Component {
...
               <input
                 type='text'
-                placeholder='Expression'
-                value={expression}
-                onChange={this.onChangeExpression.bind(this)}
-                onBlur={this.onBlurExpression.bind(this)}
+                placeholder='variable_name'
+                value={variable_name}
+                onChange={this.onChangeVariableName.bind(this)}
+                onKeyDown={(e) => {if (e.keyCode === 13 && this.isValid()) {this.onSubmit()}}}
```
mmcdermott: Not your doing, I know, but if you're editing this anyways probably best to pull this into a `submitIfEnter` helper or something (it is repeated below).

src/components/facts/list/item.js
```diff
@@ -0,0 +1,45 @@
...
+        <div className='variable-name'>
+          <span className='prefix'>#</span>
+          <span className='variable'>{fact.variable_name}</span>
+        </div>
+        <div className='description'>{fact.name}</div>
+      </div>
+
+      <div className='section-help'>
+        <span className='ui button tiny options' onClick={onEdit}>Edit</span>
+          <span className='ui button tiny options' onClick={onDelete}>Delete</span>
```
mmcdermott: Unnecessary indentation.

src/components/facts/list/container.js
OAGr: This is needed to be called like this in the space, where it doesn't have immediate access to the organization.  It's a short query.

PR #688, opened 2016-08-04
Match sub-tab via URL paths.

mmcdermott:


mmcdermott:
I've included the url changing now. I like leaving /organizations/3 an independent route, as it is not guaranteed to always be the same as /organizations/3/models.

OAGr:
That's what I meant above.  This works though.

PR #694, opened 2016-08-05
More descriptive failed fact reference error.

mmcdermott:


src/lib/propagation/metric-propagation.js
```diff
@@ -70,6 +70,7 @@ export default class MetricPropagation {
       return this.simulate(sampleCount, graph, dispatch).then(
         simulation => {
           if (simulation) {
+            console.log(simulation)
```
OAGr: should this be there?

PR #695, opened 2016-08-05
Fact input should support data input.

mmcdermott:


mmcdermott:
Test failure is illusory; tests actually pass. just codeship not reading the output correctly.

PR #696, opened 2016-08-05
Fixing metric relationship types by resolving to inputs prior to edge finding.

mmcdermott:
This is a bit hacky, and we should move this code to depend on expressions eventually.

Also, this also needs to be put into master before next deploy, but I'm hoping we can just merge this into the epic then the epic into master and sort it all in one swoop.

### DIFF
--- a/src/lib/engine/space.js
+++ b/src/lib/engine/space.js
@@ -45,20 +45,19 @@ export function toDSpace(spaceId, graph, organizationFacts) {
   if (!space) { return {} }
 
   let dSpace = Object.assign(space.asMutable(), toDgraph(space.id, graph))
-  dSpace.edges = _dGraph.dependencyMap(dSpace)
 
   const facts = possibleFacts(dSpace, graph, organizationFacts)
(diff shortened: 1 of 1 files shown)


src/lib/engine/space.js
```diff
@@ -45,20 +45,19 @@ export function toDSpace(spaceId, graph, organizationFacts) {
...
     let edges = {}
     edges.inputs = dSpace.edges.filter(i => i.output === s.id).map(e => e.input)
     edges.outputs = dSpace.edges.filter(i => i.input === s.id).map(e => e.output)
     edges.inputMetrics = edges.inputs.map(i => dSpace.metrics.find(m => m.id === i))
-    return {
-      ...s,
-      edges,
-      guesstimate: withInputFn(s.guesstimate),
-    }
+    return { ...s, edges }
```
OAGr: why did you remove ``guesstimate`` here?  Just checking to make sure that wasn't a mistake.

mmcdermott: I moved it above. Right now, we resolve edges based on the guesstimate's input and readable IDs, so the expressions need to be resolved into inputs before we compute the edges.

PR #718, opened 2016-08-09
There are only 2 's's in 'guesstimate'

mmcdermott:
This was causing all metric cards to update on any change.


OAGr:
This is kind of hilarious, kind of awful.

Issue #747, opened 2016-08-14
Specific cell should be placed into 'analysis' mode


OAGr:
This cell should be green.  This should go in after #746

PR #756, opened 2016-08-17
New sensitivity mode

OAGr:

### DIFF
--- a/src/components/lib/FlowGrid/background-container.js
+++ b/src/components/lib/FlowGrid/background-container.js
@@ -45,7 +45,7 @@ export class BackgroundContainer extends Component {
       !_.isEqual(this.props.copiedRegion, nextProps.copiedRegion) ||
       !_.isEqual(this.props.selectedRegion, nextProps.selectedRegion) ||
       !_.isEqual(this.props.autoFillRegion, nextProps.autoFillRegion) ||
-      !_.isEqual(this.props.analyzedRegion, nextState.analyzedRegion) ||
+      !_.isEqual(this.props.analyzedRegion, nextProps.analyzedRegion) ||
       !_.isEqual(this.props.edges, nextProps.edges) ||
       !_.isEqual(this.state.rowHeights, nextState.rowHeights)
     )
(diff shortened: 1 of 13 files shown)


src/components/lib/FlowGrid/background-container.js
```diff
@@ -45,46 +45,44 @@ export class BackgroundContainer extends Component {
       !_.isEqual(this.props.copiedRegion, nextProps.copiedRegion) ||
       !_.isEqual(this.props.selectedRegion, nextProps.selectedRegion) ||
       !_.isEqual(this.props.autoFillRegion, nextProps.autoFillRegion) ||
+      !_.isEqual(this.props.analyzedRegion, nextState.analyzedRegion) ||
```
mmcdermott: should this be nextProps.analyzedRegion?

OAGr: yep, good notice

src/components/lib/FlowGrid/background-container.js
```diff
@@ -45,46 +45,44 @@ export class BackgroundContainer extends Component {
...
-            rowHeights={rowHeights}
-            columnWidth={columnWidth}
-            selectedRegion={autoFillRegion}
-            type='fill'
-          />
-        }
+        {regions.map(region => {
+          if (region[0].length === 2) {
+            return (
+              <Region
```
mmcdermott: A few small things:
0. I'd rename these from regions to backgroundRegions or something to differentiate from the grid regions they contain.
1. There is a `isRegion` helper in the lcoation utils to ensure that a region is a region that you should use instead of checking the length.
2. I'd do make a `renderRegion([region, regionType])` function within this class, then structure this like: 
```
regions.map(region => isRegion(region[0]) ? this.renderRegion(region) : false)
```

src/components/metrics/card/index.js
```diff
@@ -118,6 +134,10 @@ export default class MetricCard extends Component {
     this.setState({sidebarIsOpen: (!this.state.sidebarIsOpen), modalIsOpen: false});
   }
 
+  _closeSidebar() {
+     this.state.sidebarIsOpen && this.setState({sidebarIsOpen: false})
```
mmcdermott: nit: I'd prefer to just use a normal if statement (though feel free to use a one liner) for control flow logic, I find it cleaner.

OAGr:
This would close https://github.com/getguesstimate/guesstimate-app/issues/747

Issue #744, opened 2016-08-14
Pasted links in description & text form should be converted to markdown


OAGr:
Right now many people paste in links, and these look quite bad.  They should be limited in what they show to around 40 characters or whatever when pasted in.

PR #759, opened 2016-08-18
Links should translate into markdown links upon save.

mmcdermott:


src/components/metrics/card/index.js
```diff
@@ -182,7 +184,20 @@ export default class MetricCard extends Component {
...
+    let prevIndex = 0
+    let matchArr
+    while ((matchArr = fullRegex.exec(rawDescription)) !== null) {
+      const partial = rawDescription.slice(prevIndex, matchArr.index)
+      description += partial.replace(URL_REGEX, match => `[${match}](${match})`)
+      description += matchArr[0]
+      prevIndex = matchArr.index + matchArr[0].length
+    }
+    const partial = rawDescription.slice(prevIndex)
+    description += partial.replace(URL_REGEX, match => `[${match}](${match})`)
```
OAGr: can the url changing part of this be moved into a separate function?  I imagine testing could be useful for that, but it's not a huge concern now.

OAGr:
This seems to break with hyphens, likely other characters too.  

image

OAGr:
This closes https://github.com/getguesstimate/guesstimate-app/issues/744

Issue #748, opened 2016-08-14
ReadableId should automatically adjust after name is changed


OAGr:
People in Denmark -> PiD
Names should be 2-4 characters, should be unique within the model

PR #760, opened 2016-08-18
Set the metric readable id based on the metric name. Closes 748.

mmcdermott:


OAGr:
image

Also, everything should be capitalized.

OAGr:
Better, though right now this also abridges 'Population of Norway' to 'POP'.  It seems like 'PON' would be more intuitive.  We could make that another story; sorry for noticing it at the end.  Could also discuss more in person.

mmcdermott:
Fixed. There was an existing error in the name to variable name function for facts that was preventing multi-word names being resolved to an acronym rather than an abbreviation, likely a holdout from our double-switching of directions on this section.

OAGr:
This closes https://github.com/getguesstimate/guesstimate-app/issues/748

PR #798, opened 2016-08-31
Graph propagation lib refactor.

mmcdermott:
This refactors the graph propagation library in preparation for writable facts.
Future work along this thread includes:
  * Unifying error typespace & improving internal error API.
  * Not making 'fake metrics' for facts, and instead adding them as simulation nodes directly.
  * incorporating the Guesstimator library directly into the simulation node library.
  * Additional improvements to the recorder's interactions with this library (e.g. better isolation).
  * Revise guesstimate's usage API (e.g. `graphFilters`) to better mesh with new API and reduce wrapper complexity.

This PR is deployed to matthew.getguesstimate.com, and should be smoke tested extensively.


mmcdermott:
Whoops, tests are failing. will fix and re-open.

src/lib/engine/collections.js
```diff
@@ -4,8 +4,12 @@ const equalsProp = (propValue, propName) => e => typeSafeEq(_.get(e, propName),
 const nullFn = x => null
 
 export const get = (collection, id, prop='id') => allPresent(collection, id) ? collection.find(equalsProp(id, prop)) : null
+export const gget = (collection, id, searchProp, getProp) => _.get(get(collection, id, searchProp), getProp)
```
OAGr: what does gget mean?

src/lib/engine/facts.js
```diff
@@ -94,33 +95,32 @@ export function addFactsToSpaceGraph({metrics, guesstimates, simulations}, globa
...
-export function simulateFact(fact) {
-  const fakeSelector = ['fact']
-  const guesstimate = toGuesstimate(fakeSelector, fact)
-  const graph = {metrics: [{id: guesstimate.metric}], guesstimates: [guesstimate]}
-  const metricPropagation = new MetricPropagation(guesstimate.metric, [], 0)
-
-  return metricPropagation.simulate(metricPropagation.remainingSimulations[0], graph).then(
-    ({sample: {values, errors}}) => ({values, errors})
-  )
+export function simulateFact(fact, numSamples=5000) {
```
OAGr: it seems weird to me to have the 5000 here

src/lib/engine/collections.js
```diff
@@ -4,8 +4,12 @@ const equalsProp = (propValue, propName) => e => typeSafeEq(_.get(e, propName),
...
 export const getFn = (coll, getProp='id', inProp='id') => !coll ? nullFn : e => get(coll, _.get(e, inProp), getProp)
 
 export const filter = (collection, id, prop='id') => allPresent(collection, id) ? collection.filter(equalsProp(id, prop)) : []
+export const some = (collection, id, prop='id') => allPresent(collection, id) ? _.some(collection, equalsProp(id, prop)) : false
 
 export const isPresent = e => !!e && !_.isEmpty(e)
+
+export const andFns = (...predFns) => x => predFns.reduce( (running, currFn) => running && !!currFn(x), true )
```
OAGr: Lodash may make this easier

src/lib/engine/facts.js
```diff
@@ -94,33 +95,32 @@ export function addFactsToSpaceGraph({metrics, guesstimates, simulations}, globa
...
+  const e = { text: fact.expression, guesstimateType: null }
```
OAGr: what is const name? e?  Also, not sure about 'gtr' below.

src/lib/propagation/DAG.js
```diff
@@ -0,0 +1,63 @@
...
+
+export class SimulationDAG {
+  constructor(nodes) {
+    if (!!_.get(window, 'recorder')) { window.recorder.recordSimulationDAGConstructionStart(this) }
+    let rest = nodes.map(NodeFns.extractInputs)
+
+    let byId = _.transform(rest, (map, node) => {map[node.id] = {node, lastAncestors: node.inputs, ancestors: node.inputs}}, {})
+
+    const missingInputsNodes = _.remove(rest, NodeFns.hasMissingInputs(rest))
+    let graphErrorNodes = missingInputsNodes.map(NodeFns.withMissingInputError(nodes))
```
OAGr: Above two lines confuse me.

src/lib/propagation/DAG.js
```diff
@@ -0,0 +1,63 @@
...
+    let errorNodes = Object.assign([], graphErrorNodes)
```
OAGr: does assign work with arrays?

src/lib/engine/collections.js
mmcdermott: Nope, not afaict. I checked. And it produces a function that returns true iff all the predicate functions passed in return true on that input.

src/lib/propagation/DAG.js
```diff
@@ -0,0 +1,63 @@
+import * as NodeFns from './nodeFns'
+import {SimulationNode} from './node'
+import * as constants from './constants'
+
+import * as _collections from 'gEngine/collections'
+
+export class SimulationDAG {
+  constructor(nodes) {
+    if (!!_.get(window, 'recorder')) { window.recorder.recordSimulationDAGConstructionStart(this) }
+    let rest = nodes.map(NodeFns.extractInputs)
```
OAGr: rename, maybe ``remaining_nodes`` or ``unprocessed_nodes``

src/lib/propagation/DAG.js
mmcdermott: Yes.

src/lib/engine/collections.js
mmcdermott: get from the result of a get. 

So, the way I read is, if you call:
```_collections.gget(metrics, 3, 'id', 'readableId')```
then that means "Get the property 'readableId' from the metric with 'id' '3' within the collection 'metrics'"

File: src/lib/engine/facts.js (hunk 2 above)
Comment (mmcdermott) [inline src/lib/engine/facts.js]: Yeah, it is. I can pull that out into a constant somewhere, if that'd be better. Any suggestions as to where? It was stored as an esoteric internal variable to MetricPropagation before, so doesn't exist anywhere properly atm.

File: src/lib/engine/facts.js (hunk 4 above)
Comment (mmcdermott) [inline src/lib/engine/facts.js]: `gtr` stands for `Guesstimator`. I wasn't sure what to call it, as I'm not really sure what that object is in its own right. Any suggestions?

File: src/lib/propagation/node.js [hunk 8]
```diff
@@ -0,0 +1,79 @@
...
+    this.expression = expression
+    this.type = type
+    this.guesstimateType = guesstimateType
+    this.samples = samples
+    this.errors = errors
+    this.parentIndices = parentIndices
+    this.childrenIndices = childrenIndices
+    this.ancestors = ancestors
+    this.DAG = DAG
+    this.index = index
```
OAGr: Does this need all of these?  Like the Indices?

src/lib/propagation/node.js
```diff
@@ -0,0 +1,79 @@
...
+      if (!!ancestorError) {
+        ancestorError.ancestors = _.uniq([...ancestorError.ancestors, this.id])
+      } else {
+        n.errors.push({type: GRAPH_ERROR, subType: INVALID_ANCESTOR_ERROR, ancestors: [this.id]})
+      }
+    })
+  }
+
+  hasInputErrors() { return _collections.some(this.errors, INVALID_ANCESTOR_ERROR, 'subType') }
+  getResults() { return _.pick(this, ['samples', 'errors']) }
```
OAGr: can some of these funcs be private?

src/lib/engine/facts.js
mmcdermott: what do you mean by 'const name'? I don't see that in this code.

src/lib/propagation/node.js
```diff
@@ -0,0 +1,79 @@
...
+
+  data() { return this.type === NODE_TYPES.DATA ? this.samples : [] }
+
+  parse() {
+    const e = { text: this.expression, guesstimateType: this.guesstimateType, data: this.data() }
+    const formatter = _matchingFormatter(e)
+    return [formatter.error(e), formatter.format(e)]
+  }
+
+  getInputs() {
```
OAGr: maybe you could use es6 getters and setters?
http://odetocode.com/blogs/scott/archive/2014/10/14/features-of-es6-part-9-classes.aspx

src/lib/propagation/DAG.js
mmcdermott: How so? Happy to make less opaque, but I'm not exactly sure where you're confused. What they do is first extract all nodes from the list that have missing inputs (i.e. inputs not contained in the passed in list of nodes), then add those to the list of graph error nodes.

src/lib/propagation/node.js
mmcdermott: It needs most of them. The childrenIndices aren't used presently, but may be needed later, if the graph gets too big. Should I take them out for now?

src/lib/engine/facts.js
OAGr: I mean, 'e'.  Can that be rennamed?

src/lib/propagation/node.js
OAGr: I would.  It could still be on line 17, but not the rest.

src/lib/engine/facts.js
```diff
@@ -94,33 +95,32 @@ export function addFactsToSpaceGraph({metrics, guesstimates, simulations}, globa
...
   let readableIds = metrics.map(m => m.readableId)
-  let [factMetrics, factGuesstimates, factSimulations, factHandleMap] = [[], [], [], {}]
+  let [factMetrics, factGuesstimates, factSimulations, factIdMap] = [[], [], [], {}]
   grouped.forEach(([handle, selector, fact]) => {
     // We construct virtual metrics, guesstimates, and simulations from the selector, fact, and running readable IDs...
     const {metric, guesstimate, simulation} = buildFullNode(selector, fact, readableIds)
 
     // Then update all the running variables.
-    factHandleMap[handle] = metric.readableId
+    const idToTranslate = !!_.get(fact, 'id') ? `\$\{fact:${fact.id}\}` : handle
```
OAGr: _.has instead of !!_.get

src/lib/engine/facts.js
OAGr: you could rename gtr, guesstimator, lowercase

src/lib/engine/utils.js
```diff
@@ -39,3 +39,11 @@ export function makeURLsMarkdown(text) {
 
   return transformedText
 }
+
+export const indicesOf = (list, predFn) => list.map((e, i) => predFn(e) ? i : null).filter(e => _.isFinite(e))
+export function getSubMatches(str, regex, matchIndex) {
+  if (_.isEmpty(str)) { return [] }
+  let matches = [], match
```
OAGr: I think you should declare match on a newline, just because we don't use this pattern elsewhere (from what I can tell)

src/lib/engine/utils.js
```diff
@@ -39,3 +39,11 @@ export function makeURLsMarkdown(text) {
 
   return transformedText
 }
+
+export const indicesOf = (list, predFn) => list.map((e, i) => predFn(e) ? i : null).filter(e => _.isFinite(e))
+export function getSubMatches(str, regex, matchIndex) {
```
OAGr: Documentation and/or testing would be quite nice for some of these.  At least, a small example or something.

src/lib/guesstimator/samplers/DistributionBeta.js
```diff
@@ -1,7 +1,7 @@
 import {simulate} from './Simulator.js'
 
 export var Sampler = {
-  sample({params: [hits, total]}, n) {
+  sample({params: [hits, total]}, n, _1) {
```
OAGr: what is _1?

src/lib/guesstimator/samplers/Simulator.js
```diff
@@ -83,12 +86,20 @@ function modularSlice(array, from, to) {
   return [...array.slice(newFrom), array.slice(0,to)]
 }
 
-function buildData(expr, prevModularIndex, numSamples, inputs) {
```
OAGr: realize this existed before, but the name ``buildData`` should be more specific.

src/lib/propagation/DAG.js
```diff
@@ -0,0 +1,63 @@
+import * as NodeFns from './nodeFns'
```
OAGr: I don't like the capitalization NodeFns.  Maybe nodeFns or even _nodeFncs?

src/lib/propagation/DAG.js
```diff
@@ -0,0 +1,63 @@
...
+import * as constants from './constants'
+
+import * as _collections from 'gEngine/collections'
+
+export class SimulationDAG {
+  constructor(nodes) {
+    if (!!_.get(window, 'recorder')) { window.recorder.recordSimulationDAGConstructionStart(this) }
+    let rest = nodes.map(NodeFns.extractInputs)
+
+    let byId = _.transform(rest, (map, node) => {map[node.id] = {node, lastAncestors: node.inputs, ancestors: node.inputs}}, {})
```
OAGr: i'd rename ``map`` to ``resultMap``.  Change ``byId`` to ``remaining_nodes_by_id`` or something.  Don't mind being a bit more explicit, this code now is quite terse (though obviously well thought out).

src/lib/propagation/DAG.js
OAGr: above, I'd rename ``graphErrorNodes`` to ``missingInputNodesWithErrors`` or similar, if that's what they are.  Or, ``nodesWithErrors``

OAGr: Other things: 
- missingInputsNodes shouldn't be made a const if it doesn't get used again (just move to next line or make that two lines).
- what is the diff between graphErrorNodes and errorNodes?

mmcdermott: graphErrorNodes contains more than just nodes with missing inputs; it also contains nodes with ancestral errors, or nodes with infinite loop errors. `errorNodes` is a superset of `graphErrorNodes`, containing additionally all nodes that come in with errors. These are stored separately as we extract all nodes with graph errors from the DAG prior to simulation, as they will never be able to be simulated successfully, whereas some nodes within `errorNodes` are re-simulated, in case a changed expression means the error will naturally be resolved. Also, `missingInputNodes` is used again, later, as the input to the `.map(withMissingInputErrorFn)` line below. I thought that would be a lot to put on one line. 

Unless I misunderstood your comments, I think those address your points here? Happy to make other changes, just don't want to make changes at the behest of incomplete information. I can also add more comments to this section to make things clearer.

src/lib/propagation/DAG.js
```diff
@@ -0,0 +1,63 @@
...
+      })
+    }
+
+    const withRelatives = heightOrderedNodes.map(NodeFns.withRelatives(heightOrderedNodes, byId))
+    const asNodes = withRelatives.map((n,i) => new SimulationNode(n, this, i))
+
+    this.nodes = asNodes
+    this.graphErrorNodes = graphErrorNodes
+
+    if (!!_.get(window, 'recorder')) { window.recorder.recordSimulationDAGConstructionStop(this) }
```
OAGr: I'd probably put most of the above in at least one other separate function,
``{orderedNodes, errorNodes} = sortNodes(nodes)``

src/lib/propagation/DAG.js
```diff
@@ -0,0 +1,63 @@
...
+      graphErrorNodes.push(...withInfiniteLoopErrors)
+
+      const inputErrorNodes = _.remove(
+        rest,
+        _collections.andFns(NodeFns.anyInputsWithin(errorNodes), NodeFns.allInputsWithin([...heightOrderedNodes, ...errorNodes]))
+      )
+      const withAncestralErrors = inputErrorNodes.map(NodeFns.withAncestralError(errorNodes))
+      errorNodes.push(...withAncestralErrors)
+      graphErrorNodes.push(...withAncestralErrors)
+
```
OAGr: I bet there's a cleaner way to do all of this (though I realize it is much cleaner than it could be, and this stuff is complex).  Having one object/class for the collection could make this very neat:
``DAGnodes.sortErrors``, then in ``sortErrors``: 
``{goodNodes, badNodes} = DAGnodes.checkIncomingErrors.checkInfiniteLoopErrors.checkInputErrors``

src/lib/guesstimator/samplers/DistributionBeta.js
mmcdermott: `_1` is just to indicate an unused parameter. `_` is a common convention ime for unused parameters, and javascript doesn't allow it to be used for multiple parameters, but we use that for lodash, so I use `_1`. This is useful because when that function is called, it is actually called with 3 parameters, not two, meaning if you want to add another parameter to it (as I did, then removed, when collected performance data) you need to add an unused placeholder in the third position. Naming that placeholder seems unclear to me as it is unused. 

If you don't like that convention, let me know what you'd like to do instead (either a different convention for an unused parameter, or omit it, or name the parameter despite it being unused, or whatever).

src/lib/engine/facts.js
mmcdermott: I'm not using `_.has` intentionally here, because if the fact is, for example ```
{id: null, ...}
```
then `_.has` would return true, whereas I want it to return false.

Agreed about fact being constant, but this is a transient bit of code anyways, so I'm hesitant to spend much effort cleaning it up. Do you think it is worth it, as removing this is the next feature on my list regardless?

File: src/lib/propagation/node.js (hunk 10 above)
Comment (mmcdermott) [inline src/lib/propagation/node.js]: I considered it, but didn't think it would make much sense as these are true functions more so than property accessors. Would you prefer I go that route, or was this a suggestion for consideration rather than one demonstrating opinion?

File: src/lib/engine/utils.js (hunk 13 above)
Comment (mmcdermott) [inline src/lib/engine/utils.js]: Agreed, but I think that belongs in a separate story. The changes to this file in this PR are rather minimal.

File: src/lib/propagation/DAG.js (hunk 17 above)
Comment (mmcdermott) [inline src/lib/propagation/DAG.js]: Both `lastAncestors` and `ancestors` get mutated as the iteration proceeds, and they do diverge (though they start out the same). `lastAncestors` are those farthest away from the node in question processed so far. I'm definitely open to better names there.

File: src/lib/propagation/node.js [hunk 20]
```diff
@@ -0,0 +1,77 @@
...
+    this.guesstimateType = guesstimateType
+    this.samples = samples
+    this.errors = errors
+    this.parentIndices = parentIndices
+    this.ancestors = ancestors
+    this.DAG = DAG
+    this.index = index
+  }
+
+  _data() { return this.type === NODE_TYPES.DATA ? this.samples : [] }
```
OAGr: I'd probably re-order the functions below, also keep private functions below public ones (not always, but at least with _data)

src/lib/propagation/node.js
```diff
@@ -0,0 +1,77 @@
...
+    const inputs = this._getInputs()
+
+    if (!!_.get(window, 'recorder')) { window.recorder.recordNodeSampleStart(this) }
+    const gtr = new Guesstimator({parsedError, parsedInput})
+    return gtr.sample(numSamples, inputs).then(({values, errors}) => {
+      if (!!_.get(window, 'recorder')) { window.recorder.recordNodeSampleStop(this) }
+
+      this.samples = _utils.orArr(values)
+      this.errors = _utils.orArr(errors)
+      if (!_.isEmpty(errors)) { this.addErrorToDescendants() }
```
OAGr: is it _addErrortoDescendants?

src/lib/propagation/DAG.js
mmcdermott: I've done what I can here; feel free to recommend any other changes you see fit.

src/lib/propagation/tests/DAG-test.js
```diff
@@ -0,0 +1,140 @@
...
+import * as _collections from 'gEngine/collections'
+
+// ERRORS:
+const {
+  NODE_TYPES,
+  ERROR_TYPES: {GRAPH_ERROR},
+  ERROR_SUBTYPES: {GRAPH_SUBTYPES: {MISSING_INPUT_ERROR, IN_INFINITE_LOOP, INVALID_ANCESTOR_ERROR}},
+} = constants
+
+const toExprId = id => `\$\{${toNodeId(id)}\}`
```
OAGr: This seems like it should be in a different file.  Same with other things here.

src/lib/propagation/tests/DAG-test.js
```diff
@@ -0,0 +1,140 @@
...
+
+const expectNodesToBe = (DAG, validIds, errorIds) => {
+  expect(DAG.graphErrorNodes.map(n => n.id)).to.deep.equal(errorIds.map(toNodeId))
+  expect(DAG.nodes.map(n => n.id)).to.deep.equal(validIds.map(toNodeId))
+}
+
+describe('construction', () => {
+  const simpleValidNodeList = [makeNode(1), makeNode(2, [1,3]), makeNode(3, [1])]
+  it ('Yields nodes ordered by height', () => {expectNodesToBe(new SimulationDAG(simpleValidNodeList), [1,3,2], [])})
+  it ('Correclty assignes nodes their relations', () => {
```
OAGr: Correctly assigns*

src/lib/propagation/tests/DAG-test.js
```diff
@@ -0,0 +1,140 @@
...
+    const DAG = new SimulationDAG(simpleValidNodeList)
+
+    expect(DAG.nodes.map(n => _.pick(n, ['parentIndices']))).to.deep.have.members([
```
OAGr: are the inputs parents or children?  I would have expected the inputs to be parents, but this seems like the outputs are the parents.

mmcdermott: The inputs are parents. the syntax is: `makeNode(2, [1,3])`, which means make a Node with input `node:2` whose inputs are nodes `1 and 3`. Keep in mind that the `have.members` matcher is un-ordered; i.e. I could reorder the expected list any which way and the test would still pass, so don't think they correspond to the order of the nodes above.

src/lib/engine/collections.js
mmcdermott: Offline Discussion - Add comments; otherwise OK.

PR #812, opened 2016-09-12
Working prototype of writable facts UI.

mmcdermott:
Styling TBD


src/lib/engine/utils.js
```diff
@@ -3,11 +3,15 @@ import * as _collections from './collections'
 // Source: https://gist.github.com/dperini/729294
 export const URL_REGEX = /(?:(?:https?|ftp):\/\/)?(?:\S+(?::\S*)?@)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,}))\.?)(?::\d{2,5})?(?:[\/?#]\S*)?/i
 
+export const mutableCopy = e => !!_.get(e, '__immutable_invariants_hold') && _.has(e, 'asMutable') ? e.asMutable() : e
```
OAGr: I'd probably make a separate function for ``isImmutable``, which is what I believe you are doing with the first half of that.  Also curious if this is the best way of doing that.

PR #813, opened 2016-09-12
Writable facts ui for fact cards

OAGr:
image

### DIFF
--- a/src/components/metrics/card/MetricCardViewSection/index.js
+++ b/src/components/metrics/card/MetricCardViewSection/index.js
@@ -6,7 +6,7 @@ import Histogram from 'gComponents/simulations/histogram/index'
 import MetricName from 'gComponents/metrics/card/name/index'
 import {DistributionSummary} from 'gComponents/distributions/summary/index'
 import StatTable from 'gComponents/simulations/stat_table/index'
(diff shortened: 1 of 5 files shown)


src/components/metrics/card/index.js
```diff
@@ -403,16 +402,16 @@ const MetricSidebar = ({
...
     {canBeMadeFact && !exportedAsFact &&
       <MetricSidebarItem
-        icon={<Icon name='bookmark'/>}
-        name={'Make Fact'}
+        icon={<i className='ion-ios-redo'/>}
+        name={'Export'}
         onClick={onMakeFact}
       />
     }
     {canBeMadeFact && exportedAsFact &&
```
mmcdermott: Delete this option, right?

src/components/metrics/card/MetricCardViewSection/index.js
```diff
@@ -113,14 +114,14 @@ export class MetricCardViewSection extends Component {
         }
 
         <div className='MetricTokenSection'>
```
mmcdermott: I would make a separate 'showMetricToken' boolean and use that to filter the rendering of this component, instead of dumping all of that logic in the sub-component.

PR #821, opened 2016-09-14
Separate sidebar facts into inputs & outputs

OAGr:


src/components/facts/list/item.js
```diff
@@ -2,43 +2,54 @@ import React, {Component} from 'react'
...
-          <span className='prefix'>#</span>
-          <div className='name'>{fact.variable_name}</div>
         </div>
-      </div>
 
         <div className='section-help'>
           <span className='ui button small options' onClick={onEdit}>Edit</span>
         </div>
+
+      {!!fact.exported_from_id &&
```
mmcdermott: What is the difference between isExportedFromSpace and !!fact.exported_from_id? And are all these different pieces of defensivity necessary?

mmcdermott: ... Ok, I think I understand the difference now, but the naming here is very confusing. Can you clarify the naming here?

src/components/facts/list/list.js
```diff
@@ -61,22 +64,41 @@ export class FactList extends Component {
     />
   }
 
-  renderFacts() {
-    const {props: {facts}, state: {editingFactId}} = this
-    if (!editingFactId) { return _.map(facts, this.renderFactShow.bind(this)) }
+  renderSpaceFacts() {
+    const {props: {facts, spaceId, imported_fact_ids}, state: {editingFactId}} = this
+    const isExportedFromSpace = fact => fact.exported_from_id === this.props.spaceId
+    const isImportedFromSpace = fact => imported_fact_ids.indexOf(fact.id) > -1
```
mmcdermott: Do you want to suffix these with `-Fn` or something, as we've done (occasionally) previously? I don't care either way, it is up to you.

src/components/facts/list/list.js
```diff
@@ -61,22 +64,41 @@ export class FactList extends Component {
...
 
-    const editingFactIndex = facts.findIndex(fact => fact.id === editingFactId)
-    return [
-      ..._.map(facts.slice(0, editingFactIndex), this.renderFactShow.bind(this)),
-      this.renderEditForm(facts[editingFactIndex]),
-      ..._.map(facts.slice(editingFactIndex + 1), this.renderFactShow.bind(this)),
-    ]
+    let filteredFacts = utils.mutableCopy(facts)
+    const exported = _.remove(filteredFacts, e => isExportedFromSpace(e))
```
mmcdermott: Don't need the `e => ...(e)`, can just use the function name.

src/components/facts/list/list.js
```diff
@@ -29,11 +30,13 @@ export class FactList extends Component {
   }
 
   renderFactShow(fact) {
+    const isExportedFromSpace = (fact.exported_from_id === this.props.spaceId)
```
mmcdermott: Do you want to make pull this function into somewhere shared as it is used twice?

src/components/spaces/show/fact_sidebar.js
```diff
@@ -0,0 +1,35 @@
...
+
+function mapStateToProps(state) {
+  return {
+    organizations: state.organizations,
+    organizationFacts: state.facts.organizationFacts,
+  }
+}
+
+@connect(mapStateToProps)
+export class FactListContainer extends Component{
```
mmcdermott: Could this be a functional component instead of a class? Also, we'll want to be wary about re-renders of this component as it contains many histograms and is thus expensive to render.

src/components/spaces/show/fact_sidebar.js
```diff
@@ -0,0 +1,35 @@
+import React, {Component, PropTypes} from 'react'
```
mmcdermott: Is this file actually used? It looks like you don't import this new component below...

src/components/facts/list/list.js
```diff
@@ -8,6 +8,9 @@ import {utils} from 'gEngine/engine'
 
 import './style.css'
 
+const isExportedFromSelectedSpaceFn = (fact, spaceId) => (fact.exported_from_id === spaceId)
+const isImportedFromSelectedSpaceFn = (fact, imported_fact_ids) => imported_fact_ids.indexOf(fact.id) > -1
```
mmcdermott: `ImportedTo`, perhaps? Also, I believe you can just do 
```
({id}, imported_fact_ids) => imported_fact_ids.includes(id)
```

mmcdermott: I find these functions really strange... Are they intended to be general purpose functions, used elsewhere, or functions specific to the components below? If specific to the components below, why not bind them to those components and have them reference the props directly? This seems like some kind of odd hybridization.

src/components/facts/list/item.js
```diff
@@ -2,43 +2,55 @@ import React, {Component} from 'react'
...
-          </div>
-         }
-        {_.has(fact, 'simulation.sample.values.length') &&
-          <div className='histogram'>
-            <Histogram
-              height={15}
-              simulation={fact.simulation}
-              cutOffRatio={0.995}
+export const FactItem = ({fact, onEdit, isExportedFromSelectedSpace}) => {
+  const exported_from_url = `${spaceUrlById(_.get(fact, 'exported_from_id'))}?factsShown=true`
```
mmcdermott: This should probably be an engine function. NBD for now though.

src/components/facts/list/item.js
```diff
@@ -2,43 +2,55 @@ import React, {Component} from 'react'
...
+  return (
+    <div className='Fact--outer'>
+      <div className='Fact'>
+        <div className='section-simulation simulation-sample'>
+          {_.has(fact, 'simulation.sample.values.length') && _.has(fact, 'simulation.stats.mean') &&
+            <div className='simulation-summary'>
+            <DistributionSummary
+              length={fact.simulation.sample.values.length}
```
mmcdermott: We should also make simulation engine functions for these accessors. We have some for values and errors, it would be good to do the same for length, mean, and the rest. It would help buffer us against any data structure changes down the line. No need to do anything in this PR, thought.

PR #848, opened 2016-09-24
Added new fact-auto-generate methods and tests to improve generated fact variable names. Closes 832.

mmcdermott:


src/lib/generateVariableNames/nameToVariableName.js
```diff
@@ -8,38 +12,69 @@ function prepareName(rawName) {
...
+        const lastWord = words[numberOfWordsUsed]
+        if (!BAD_ENDING_WORDS.includes(lastWord)) { break }
+      }
+
+      return _.trimEnd(words.slice(0, numberOfWordsUsed+1).join('_'), '_')
+    }
   }
 }
 
 export function getVariableNameFromName(
```
OAGr: This is a huge list of ordered params, can it be changed to a hash?

src/lib/generateVariableNames/tests/generateFactVariableName-test.js
```diff
@@ -0,0 +1,130 @@
...
+    {
+      description: `
+        A fact with a name with only spaces should not yield a new variable name.
+      `,
+      fact: {name: '     '},
+      shouldGenerateVariableName: false,
+    },
+    {
+      description: `
+        A fact with a name with only spaces and digits should not yield a new variable name.
```
OAGr: Question: What are we supposed to do in this case?  Users do need to use var names to use facts.

src/lib/generateVariableNames/tests/generateFactVariableName-test.js
```diff
@@ -0,0 +1,130 @@
...
+    {
+      description: `
+        A fact with a name with only foreign characters should not yield a new variable name.
+      `,
+      fact: {name: 'Люди в Бостоне, מענטשן אין באָסטאָן, बोस्टन में लोग, مردم در بوستو.'},
+      shouldGenerateVariableName: false,
+    },
+    {
+      description: `
+        A fact with a name with a leading number should drop the number in the variable name.
```
OAGr: why?

src/lib/generateVariableNames/tests/generateFactVariableName-test.js
mmcdermott: The will have to assign it manually. That's still an option on the form.

src/lib/generateVariableNames/tests/generateFactVariableName-test.js
mmcdermott: Honestly, mostly because metrics required that. I'd rather change that restriction later, though, when we more fully separate the metric readableId generation from the fact variable name generation.

Also, its a bit awkward not assigning anything for something that is only digits and assigning something for something that is only digits followed by a word.

src/lib/generateVariableNames/tests/generateFactVariableName-test.js
OAGr: Ok, that makes sense for now.

src/lib/generateVariableNames/tests/generateFactVariableName-test.js
OAGr: That wrks.

PR #854, opened 2016-09-25
Add necessary props to metric Modal to enable organizational facts.

mmcdermott:


mmcdermott:
Closing until #855 is better understood.

PR #868, opened 2016-09-26
Metrics should only be exportable once. Closes 856.

mmcdermott:


OAGr:
Closes https://github.com/getguesstimate/guesstimate-app/pull/868

PR #869, opened 2016-09-26
Orphan facts. Closes 828.

mmcdermott:


src/components/organizations/show/facts/factGraph.js
```diff
@@ -140,7 +142,13 @@ export class FactGraph extends Component {
       edges.push(...parents.map(p => ({input: locationById(p), output: locationById(id), pathStatus})))
     })
 
-    return { items, edges: edges.filter(e => !!e.input && !!e.output) }
+    const bad_edges = _.remove(edges, ({input, output}) => !_utils.allPresent(input.row, input.column, output.row, output.column))
+    if (!_.isEmpty(bad_edges)) {
+      console.warn(bad_edges.legnth, 'BAD EDGES ENCOUNTERED!')
```
OAGr: length*

src/components/organizations/show/facts/factGraph.js
```diff
@@ -72,6 +72,8 @@ export class FactGraph extends Component {
       component: <SpaceCard space={s} />,
     }))
 
+    const orphanedFactNodes = _.remove(factNodes, _.negate(allParentsWithin(spaceNodes)))
```
OAGr: leave comment to explain this.

src/modules/facts/reducer.js
```diff
@@ -108,6 +108,23 @@ export function factsR(state = INITIAL_STATE, action) {
         ],
       }
     }
+    case 'SPACES_DELETE_SUCCESS': {
```
OAGr: why is this necessary for this PR?

src/modules/metrics/actions.js
```diff
@@ -22,11 +21,23 @@ export function addMetric(item) {
   }
 }
 
-//spaceId must be done before the metric is removed here.
 export function removeMetrics(ids) {
```
OAGr: any way this could happen on the server?  I realize that would be tricky but it does seem more right to me.

src/modules/metrics/actions.js
```diff
@@ -22,11 +21,23 @@ export function addMetric(item) {
...
+
     const spaceId = findSpaceId(getState, ids[0])
+
+    const {organizations, spaces, facts: {organizationFacts}} = getState()
+    const organizationId = e.collections.gget(spaces, spaceId, 'id', 'organization_id')
+    if (!!organizationId) {
+      const organization = e.collections.get(organizations, organizationId)
+      const facts = e.organization.findFacts(organizationId, organizationFacts)
+      const factsToDelete = _.filter(facts, f => f.exported_from_id === spaceId && ids.includes(f.metric_id))
+      factsToDelete.forEach(fact => {
```
OAGr: It would be nicer if they could be deleted as part of one action, though i realize that practically speaking one metric removal should only go to one fact removal.

mmcdermott: I agree, but soon I'd like to refactor the facts client side storage (get rid of the organizationFacts middle layer and just store facts by ID like everything else) and this section will undergo heavy changes then that will make it more amenable to such a change, so I'd prefer to make that change then.

src/modules/metrics/actions.js
mmcdermott: We talked offline, decided to leave this here for now.

PR #871, opened 2016-09-27
Allow user to make fact on space page

OAGr:


src/components/spaces/show/index.js
```diff
@@ -330,7 +330,7 @@ export default class SpacesShow extends Component {
                 existingVariableNames={organizationFacts.map(e.facts.getVar)}
                 facts={organizationFacts}
                 organization={organization}
-                isEditable={false}
+                isEditable={true}
```
mmcdermott: This doesn't actually seem to prevent editing of the individual facts BTW. No big deal for now, though.

PR #878, opened 2016-09-28
Do not allow adjusted interval to break

OAGr:
The big issue was that if an array had all of one element, ``meanIndex`` would be -1, which would really mess up the inputs of the percentile functions in adjustedLow and adjustedHigh.


mmcdermott:
Try this instead; I'm not sure the way you've done this here is the best way. https://github.com/getguesstimate/guesstimate-app/pull/879

PR #890, opened 2016-10-06
Error message refactors.

mmcdermott:
This breaks some tests, which are really broken because the graph construction is itself flakey (sensitive to node ordering, bad with incoming errors, all true in prod as well as here), which I will fix in a follow up PR.


src/components/metrics/card/MetricCardViewSection/index.js
```diff
@@ -84,6 +83,24 @@ export class MetricCardViewSection extends Component {
     }
   }
 
+  renderErrorSection() {
+    const { metric: {name}, idMap, inSelectedCell, hovered } = this.props
+
+    if (!(this._hasErrors() && !inSelectedCell)) { return false }
```
OAGr: simplify that statement

src/components/metrics/card/MetricCardViewSection/index.js
```diff
@@ -84,6 +83,24 @@ export class MetricCardViewSection extends Component {
...
+
+    const nodeIdMap = _.transform(idMap, (res, value, key) => {res[metricIdToNodeId(key)] = value}, {})
```
OAGr: what does 'res' stand for?

PR #907, opened 2016-10-15
Fix no uncertainty display.

mmcdermott:
Bit worried about perf with this, but let me know what you think on it regardless. Let's not merge it though until we have more data.


OAGr:
I'm ok with this methodology, will wait on you for the perf data.

PR #910, opened 2016-10-18
Show metric samples

OAGr:
image


src/components/metrics/modal/index.js
```diff
@@ -17,6 +18,18 @@ const percentages = (values, perc) => {
   return perc.map(e => { return {percentage: e, value: percentile(values, values.length, e)} })
 }
 
+const SampleList = ({samples}) => (
+  <ul className='SampleList'>
+    {_.map(samples, (element, index) => {
+      return (
+        <li key={index}>
+          <div key={index}>{element}</div>
```
mmcdermott: Mostly just for my own curiosity, why does the `div` need a `key` prop?

src/components/metrics/modal/index.js
```diff
@@ -109,6 +123,21 @@ export class MetricModal extends Component {
...
               </div>
+              <div className='col-sm-2'>
+
+              <div className='metricModal--sample-container'>
+                <DropDown
+                  headerText={'Samples'}
+                  openLink={<a className='modal-action button'> Samples </a>}
+                  position='right'
+                  hasPadding={true}
+                  ref='DropDown'
```
mmcdermott: Tiny thing, but can we not add a `ref` here? Sometimes `ref`s cause problems and this one doesn't appear used, is all.

src/components/metrics/modal/index.js
OAGr: I believe React generally does for lists, I copied this from the data field where this was required.

src/components/metrics/modal/index.js
OAGr: Will do.

PR #911, opened 2016-10-18
Highlight first degree edges differently

OAGr:
This colors 1st-degree edges differently from other edges.

image

image

https://productpains.com/post/guesstimate/style-direct-sibling-edges-of-nodes-with-brighter-color

Model credit: https://www.getguesstimate.com/models/1307

### DIFF
--- a/src/components/lib/FlowGrid/edge.js
+++ b/src/components/lib/FlowGrid/edge.js
@@ -95,16 +95,17 @@ export default class Edge extends Component{
   }
 
   render() {
-    const {output, input, hasErrors, pathStatus} = this.props
+    const {output, input, hasErrors, pathStatus, className} = this.props
     if (!this._isValidNode(input) || !this._isValidNode(output)) { return (false) }
     let inputPoints = (new Rectangle(input)).showPosition(output)
     let outputPoints = (new Rectangle(output)).showPosition(input)
 
     let points = `M${inputPoints.x},${inputPoints.y} L${outputPoints.x} ,${outputPoints.y}`
 
+    console.log(className, this.props)
     return (
         <path
-            className={`basic-arrow ${pathStatus}${hasErrors ? ' hasErrors' : ''}`}
+            className={`basic-arrow ${pathStatus}${' ' + className}${hasErrors ? ' hasErrors' : ''}`}
             d={points}
             markerEnd={`url(#MarkerArrow-${hasErrors ? 'hasErrors' : pathStatus})`}
             fill="none"
--- a/src/components/spaces/canvas/index.js
+++ b/src/components/spaces/canvas/index.js
@@ -222,10 +222,16 @@ export default class Canvas extends Component{
 
       let pathStatus = unconnectedStatus
       if (!withinSelectedRegion && (outputIsAncestor || inputIsDescendant)) {
+        const degree = outputIsAncestor ?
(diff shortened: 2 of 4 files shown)


src/components/lib/FlowGrid/edge.js
```diff
@@ -95,16 +95,17 @@ export default class Edge extends Component{
...
   render() {
-    const {output, input, hasErrors, pathStatus} = this.props
+    const {output, input, hasErrors, pathStatus, className} = this.props
     if (!this._isValidNode(input) || !this._isValidNode(output)) { return (false) }
     let inputPoints = (new Rectangle(input)).showPosition(output)
     let outputPoints = (new Rectangle(output)).showPosition(input)
 
     let points = `M${inputPoints.x},${inputPoints.y} L${outputPoints.x} ,${outputPoints.y}`
 
+    console.log(className, this.props)
```
mmcdermott: This should go, right?

src/components/lib/FlowGrid/edge.js
```diff
@@ -95,16 +95,17 @@ export default class Edge extends Component{
...
     return (
         <path
-            className={`basic-arrow ${pathStatus}${hasErrors ? ' hasErrors' : ''}`}
+            className={`basic-arrow ${pathStatus}${' ' + className}${hasErrors ? ' hasErrors' : ''}`}
```
mmcdermott: Check out the `getClassName` helper in utils to clean this section up. You can do something like:
```
className={getClassName('basic-arrow', className, hasErrors ? 'hasErrors' : null)}
```

src/components/spaces/canvas/index.js
```diff
@@ -222,10 +222,16 @@ export default class Canvas extends Component{
...
+          ancestors.findIndex(e => e.id === outputId) + 1
+          :
+          descendants.findIndex(e => e.id === inputId) + 1
+
         pathStatus = outputIsAncestor ? 'ancestor' : 'descendant'
+        if (degree === 1) { pathStatus += '-1-degree' }
       }
 
-      return {input, output, pathStatus, hasErrors: hasErrors(simulation) }
+      return {input, output, pathStatus, hasErrors: hasErrors(simulation), className: '' }
```
mmcdermott: Seems like you added code to deal with you setting a `className` in other files, but then always set it to `''`? Am I missing something, or can some of this stuff go?

src/components/spaces/canvas/index.js
```diff
@@ -222,10 +222,16 @@ export default class Canvas extends Component{
 
       let pathStatus = unconnectedStatus
       if (!withinSelectedRegion && (outputIsAncestor || inputIsDescendant)) {
+        const degree = outputIsAncestor ?
+          ancestors.findIndex(e => e.id === outputId) + 1
```
mmcdermott: I'm confused by this... will come over and discuss in person.

mmcdermott: So, after looking at this, I think this will only work for when you have just one metric selected, and it works in that case because then ancestors[0] === descendants[0] === the one selected metric So this loop cycles through and catches all the edges that either lead into the selected metric as degree zero or that lead out of the selected metric as degree zero. 

I think what you really want here is something like:
```
const isDegreeOne = _.some(selectedMetrics, ({id}) => [inputId, outputId].includes(id))
```

PR #930, opened 2016-11-14
Ensuring that post-drop handlers are only called if the drop is valid. Closes 908.

mmcdermott:
Plus a few tiny refactors.


mmcdermott:
Staged to stage.getguesstimate.com

PR #935, opened 2016-11-15
Fix sub card

OAGr:


OAGr:
Go stack overflow:
http://stackoverflow.com/questions/7432747/superscript-not-working-in-a-web-browser

PR #937, opened 2016-11-16
Preventing clicks on the lognormal icon when lognormal is an invalid choice. Closes #553.

### LINKED ISSUE #553: You shouldn't be able to select the lognormal icon with an impossible lognormal range.
right now, if you type 0 to 10, you can select the lognormal icon even though that is an invalid option. Nothing breaks, it just does nothing, but should still be fixed.

mmcdermott:
And minor refactors.

### DIFF
--- a/src/components/distributions/editor/TextForm/TextForm.js
+++ b/src/components/distributions/editor/TextForm/TextForm.js
@@ -2,7 +2,9 @@ import React, {Component, PropTypes} from 'react'
 
 import {GuesstimateTypeIcon} from './GuesstimateTypeIcon'
 import {TextInput} from './TextInput'
-import DistributionSelector from './DistributionSelector'
+import {DistributionSelector} from './DistributionSelector'
+
+import {Guesstimator} from 'lib/guesstimator/index'
 
 export class TextForm extends Component{
   displayName: 'GuesstimateInputForm'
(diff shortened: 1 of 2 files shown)


src/components/distributions/editor/TextForm/TextForm.js
```diff
@@ -63,6 +68,13 @@ export class TextForm extends Component{
       m => !!_.get(m, 'simulation.sample.errors.length') || !_.get(m, 'simulation.sample.values.length')
     ).map(m => m.readableId)
 
+    // To see if this guesstimate is a valid choice for a lognormal distribution, we'll try to parse it with
+    // guesstimateType manually set to 'LOGNORMAL', and see if the parser corrects that type to something else. This
+    // approach is a bit hacky, but it gets the job done.
+    const [_1, parsed] = Guesstimator.parse({input, guesstimateType: 'LOGNORMAL'})
+    const parsedType = _.get(parsed, 'parsedInput.guesstimateType')
+    const isLognormalValid = parsedType === 'LOGNORMAL'
```
OAGr: could this be wrapped in a function, ``isLognormal`` in the engine?

mmcdermott: hmm, I'd kind of rather leave it here, and just put it into the Guesstimator/Parser later when we refactor those. This is a definite hack right now, so I'd rather not cement it in a library if that makes sense.

PR #943, opened 2016-12-13
583 flash input when inserting other readable

OAGr:


mmcdermott:
Hmm, playing with it on ozzie.getguesstimate.com, this seems a very harsh transition, and it seems like something is wrong with differentiating valid flashing inputs from invalid flashing inputs when you are inserting a metric into a function for the first time (the first occurrence of the inserted metric, not the first metric inserted overall), as the metric readable id flashes red first then turns green. Is that behavior desired? I certainly would have expected the transition to be more gradual, and the readable id to not flash red if the metric is a valid metric, though, this may be an issue with the list of valid metrics not being complete at the first render more than anything else.

OAGr:
Right now the box flashes green whenever something is added, even if the thing added produces an error. Is this the trouble? It seems to work ok for me.

OAGr:
I see the problem. The same problem was actually there in your version, before I edited, it seems. I think it has something to do with updateDecorators() being called before it's ready; ``this.props.validInputs`` and ``flashingInput`` seem to sometimes have the wrong values at the wrong times. Can you look into that?

mmcdermott:
Yeah, will do, that makes sense. I'm going to close this in the meantime.

PR #944, opened 2016-12-13
941 show pictures of org editors

OAGr:
This shows the org editor images on the spaces themselves, not that the cards.

@mmcdermott do you know if the data is always available in the cards?

### DIFF
--- a/src/lib/engine/space.js
+++ b/src/lib/engine/space.js
@@ -73,10 +73,19 @@ export function toDSpace(spaceId, graph, organizationFacts) {
 
 function toDgraph(space, graph){
   const {users, organizations, calculators, userOrganizationMemberships, me} = graph
-  const {user_id, organization_id} = space
+  const {user_id, organization_id, author_contributions} = space
+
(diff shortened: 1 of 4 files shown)


OAGr:
Could we merge this now, then later add the functionality in cards?

src/lib/engine/space.js
```diff
@@ -73,10 +73,19 @@ export function toDSpace(spaceId, graph, organizationFacts) {
 
 function toDgraph(space, graph){
   const {users, organizations, calculators, userOrganizationMemberships, me} = graph
-  const {user_id, organization_id} = space
+  const {user_id, organization_id, author_contributions} = space
+
+  let org_users = []
+
+  if (!!organization_id && !_.isEmpty(author_contributions)){
+    org_users = Object.keys(author_contributions).map(key => ({..._collections.get(users, key), edits: author_contributions[key]}))
```
mmcdermott: I would rename `key` to `author_id` or something here, I think that would be clearer.

PR #946, opened 2016-12-14
Interactive histogram range

OAGr:
recording

### DIFF
--- a/src/components/lib/histogram.js
+++ b/src/components/lib/histogram.js
@@ -3,6 +3,7 @@ import React, {Component, PropTypes} from 'react'
 import d3 from 'd3'
 
 import numberShow from 'lib/numberShower/numberShower'
+import $ from 'jquery'
 
 function getYScale(data, height) {
   return d3.scale.linear().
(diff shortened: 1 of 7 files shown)


src/components/lib/histogram.js
```diff
@@ -86,24 +87,61 @@ export default class Histogram extends Component {
...
 
-  render() {
-    let { top, right, bottom, left, data, width, height, cutOffRatio } = this.props;
+  state = {
+    xScale: (e) => e,
+    yScale: (e) => e,
+    histogramData: []
+  }
+
+  componentWillReceiveProps(nextProps){
```
mmcdermott: Missing spaces before your `{`s here and below. You can do it with a find replace I think, though don't remember the exact regex off hand.

src/components/lib/histogram.js
```diff
@@ -86,24 +87,61 @@ export default class Histogram extends Component {
...
+    if (this.needsNewScales(nextProps)){
+      this.computeNewScales(nextProps)
+    }
+  }
+
+  componentWillMount(){
```
mmcdermott: Delete this empty method now?

src/components/simulations/histogram_with_stats/index.js
```diff
@@ -0,0 +1,99 @@
...
+import $ from 'jquery'
+import './style.css'
+
+const MARGIN_LEFT = 10
+
+const percentages = (values, perc) => {
+  return perc.map(e => { return {percentage: e, value: percentile(values, values.length, e)} })
+}
+
+const findPercentile = (values, value) => {
```
mmcdermott: I think you can just use the `cutoff` function in `src/lib/dataAnalysis.js` to find the index.

PR #960, opened 2017-02-27
Bug fixes

mmcdermott:
Fixes the PERT distribution, a typo in the Sortable component that messes up fresh installs (perhaps they updated without doing a version bump, as old mechanisms no longer work at fully specified version in package.json?), and a bug with the calculator form close button being rendered beneath other visible components, and thus inaccessible.

On stage.getguesstimate.com


OAGr:
Good work

PR #989, opened 2018-08-21
Add Travis CI to run linting and tests

wdoug:
This should hopefully make it easier to review changes, knowing that tests and linting have at least passed, in addition to making those a lot more useful.

There are several options for free CI systems for open source apps. I have enjoyed working with Travis in the past and would recommend it. If some other system is wanted that is fine too, it just seems like some sort of CI system would be beneficial.

Summary of changes:
- Created package-lock file for current versions of npm
- Integrated an updated version of eslint into the repo
- Fixed some eslint errors
- Added config to run linting and tests in travis ci

The output run against my fork can be seen [here](https://travis-ci.org/wdoug/guesstimate-app/builds/418506559)

After this change, the only thing required to set up travis would be to enable this for the main repo. Travis is free for open source apps.


.eslintrc
```diff
@@ -38,9 +37,8 @@
         "react/no-unknown-property": 1,
         "react/prop-types": 1,
         "react/react-in-jsx-scope": 1,
-        "react/require-extension": 1,
```
wdoug: Unlike the other eslint rules here that were simply renamed, this one was changed to a different plugin ([see here](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md)). It would be great to add that plugin and configure some of its other rules as well, but I thought that might be more appropriate for a different pull request.

OAGr:
Thanks! I had a bit of difficulty merging this straight in, so I made a new branch and cherry-picked the working commits, and tried to adapt the other code. See: https://github.com/getguesstimate/guesstimate-app/pull/1000

PR #998, opened 2019-03-17
Add ESLint as a development dependency; update config

patbl:
This helps if you've globally installed ESLint with a different version than the version that was used when the config was originally written. (Some of the rules have changed.)


OAGr:
Looks good to me, good work!

PR #1018, opened 2022-07-19
Dependabot happy

quinn-dougherty:
- [x] Closes #1015 
- [x] adds `yarn ci:test` to github ci (I'm unaware of it still running in travis) 
- [x] Close #1014 
- [x] Close #1017 
- [x] Close #1016 
- [x] Close https://github.com/getguesstimate/guesstimate-app/pull/1019

PR #1020, opened 2022-07-28
[WILL DEPLOY TO PROD] sync production and staging

quinn-dougherty:
- #1018 reduced security alerts by 70%ish 
- #1012 added 14 security alerts


quinn-dougherty:
I'm not sure it would deploy if we merged. netlify is acting funny.

Issue #1011, opened 2022-02-24
Frontend sometimes sends auth_id=undefined, resulting in 500


Hazelfire:
I've found that the frontend sometimes sends 
`GET https://guesstimate.herokuapp.com/users?auth0_id=undefined`
The backend promptly returns a 500, Internal Server Error.

 This clearly seems to be a problem with both the backend and the frontend, in that the backend is giving a 500 on an invalid input and the frontend is creating it. I'm going to see if I can look at the heroku logs.

Hazelfire:
Ok, it seems to pass the id to auth0 without validation. Despite it having {"statusCode": 400 ... } in the error, this is a 500 error
![image](https://user-images.githubusercontent.com/13807753/155443252-4e8e3146-71af-4f73-b9d7-f4d35e5366a8.png)

Hazelfire:
Ok, that's odd, I managed to get a similar error even when there actually was an auth id...

PR #1024, opened 2022-12-22
NextJS and other improvements

berekuk:
Seems like this will to be a typical epic branch with a mix of many big refactorings.

- upgraded: react 18, modern webpack with nextjs, modern redux and some redux-toolkit, etc.
- merged all css in legacyStyles.css (necessary for upgrading since nextjs doesn't allow component-level css imports)
- added some typescript (non-strict and very lenient for now)
- had to replace most of `@connect` decorators with a more verbose syntax since decorator syntax has changed (react-redux recomments to use hooks these days)
- added prettier

This branch is still unstable and unfinished. I'm unsure which parts are unstable because of existing bugs (e.g. because of #1011) and which parts were broken by the upgrade. I'm still in the process of figuring out how everything works.

Stabilizing this PR is the first priority, but while I'm jumping around the codebase and poking my dev frontend instance, I might do some additional refactorings when it matches the current frontend best practices (e.g. functional components instead of classes), or when the change is safe and local enough.

I promise not to do any significant product-level changes here, though. (Might throw in some minor stuff that's relevant for debugging, e.g. [nprogress](https://ricostacruz.com/nprogress/) or better error reporting).


Incognito:
I don't maintain this project but...
Just a thought: Prettier makes no factual code changes but updates several hundred files, making the meaningful diffs harder for humans to find and read. Could easily be done in stand-alone PR first.

berekuk:
@Incognito I agree in theory, but in practice this doesn't seem to matter in this specific case:
- prettier commit is separate (8461ac66f4db4e30eae8dc96ea150b23a1944948) so it's still possible to review other changes with `git diff` from the specific commit hash
- the final diff for this PR will be huge anyway, and tbh I don't expect anyone to have enough time to review it all
- clear separate PRs are nice, but velocity is important too, and there's a ton of work to do here (this project was pretty unmaintained for the last few years), and I'm cutting some corners intentionally, so that this will take a week of my work instead of a month

OAGr:
@Incognito I just wanted to say, thanks for the comment, and we are really appreciative of suggestions. @berekuk is the one who makes calls here, but advice and opinions are a appreciated (even when we don't go through with them).

berekuk:
Superseded by #1035.

